PR #57, opened 2019-06-13
Fix: Install dependencies for Mac OS

selvan:
Able to make successful Mac OS build after the fix.


giongto35:
Thanks very much :D

PR #67, opened 2019-08-14
 Introduce Makefile and go vendor directory for faster build

sadlil:
1. Makefile currently includes the equivalent of `run_local.sh` and `run_local_docker.sh`
Use
```
$ make run # to run
$ make run-docker # to run docker
```
2. Include vendor directories to firster build time. Also, this introduces to consistent builds over commits. Run `make dep` to get new dependencies into the vendor.


giongto35:
Wow @_@. I didn't know about go mod vendor. Thank you very much for introducing this change.
I'm sorry that I made a change in readme and this branch is not mergeable. Can you update the branch?

sadlil:
@giongto35 Sure. Already updated.

PR #100, opened 2019-10-08
Fix GBA internal resolution scaling

sergystepanov:
Quick fix for GBA tearing. (Resolution scaling needs to be rewritten)

Before
image

After
image

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -47,8 +47,8 @@ type EmulatorMeta struct {
 var EmulatorConfig = map[string]EmulatorMeta{
 	"gba": {
 		Path:   "assets/emulator/libretro/cores/mgba_libretro.so",
-		Width:  240,
-		Height: 160,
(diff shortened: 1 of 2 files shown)


giongto35:
Yes, I also think scale down resolution will help. it is affecting resolution for other games but seems like no other way.

The cause currently is the encoding image size has to be fixed with the size declared in config. 
https://github.com/giongto35/cloud-game/blob/master/pkg/worker/room/room.go#L104

However, image output size from the game emulator fluctuates over time. 
https://github.com/giongto35/cloud-game/blob/master/pkg/emulator/libretro/nanoarch/nanoarch.go#L125
you can output "width" and "height" to see the output is changed frequently.

That's why I have to fix the size in Config beforehand.

giongto35:
I has to test this change with other GBA first :p

giongto35:
oh I see. nice catch :D

PR #107, opened 2019-10-12
Add frame scaling support

sergystepanov:
There are three types of scaling:

- noop
- nearest neighbour
- old implementation (nearest neighbour without interpolation (:)


pkg/emulator/libretro/image/scale.go
```diff
@@ -0,0 +1,80 @@
...
+	ScaleOld = 0
+	// nearest neighbour interpolation
+	ScaleNearestNeighbour = 1
+)
+
+func Resize(scaleType int, fn Format, w int, h int, packedW int, vw int, vh int, bpp int, data []byte, image *image.RGBA) {
+	// !to do set it once instead switching on each iteration
+	// !to do skip resize if w=vw h=vh
+	switch scaleType {
+	case ScaleSkip:
```
giongto35: If you want to make it looks Go. You can have an interface with resize signature. 
type resizer interface {
  func resize(fn, w, h, packedW, vw, vh, bpp, data, image)
}

giongto35:
👍 Looks really good

PR #109, opened 2019-10-14
Add and enable bilinear scaling

sergystepanov:


sergystepanov:
Maybe it's worth to port [sharp-bilinear](https://github.com/libretro/common-shaders/blob/master/interpolation/shaders/sharp-bilinear.cg) [shaders](https://github.com/rsn8887/Sharp-Bilinear-Shaders).

sergystepanov:
There is some color blending artifacts (Mario start screen, under the mountain, green is bleeding onto the rocks)
image

pkg/worker/config/config.go
```diff
@@ -1,4 +1,4 @@
-package worker
+package workerConfig
```
sadlil: I am on the side of keeping the worker's config in the same package. If worker config going to need some other shared dependecy I would suggest to move them to a third party package ie. `pkg/monitoring`

cmd/overworker/main.go
```diff
@@ -16,7 +17,7 @@ import (
 func main() {
 	rand.Seed(time.Now().UTC().UnixNano())
 
-	cfg := worker.NewDefaultConfig()
+	cfg := workerConfig.NewDefaultConfig()
```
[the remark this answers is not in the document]
sergystepanov: The main reason: I couldn't import it in room.go because:
```
import cycle not allowed
package github.com/giongto35/cloud-game/cmd/overworker
        imports github.com/giongto35/cloud-game/pkg/worker
        imports github.com/giongto35/cloud-game/pkg/worker/room
        imports github.com/giongto35/cloud-game/pkg/worker
```
Is it possible to fix this somehow?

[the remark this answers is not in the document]
sadlil: yeah. I understand the scenario. There is two way of solving this - 1. that what you did, 2. Move it to a root directory maybe something similar to - `configs/worker`. 
I don't mind having any approach. But let's keep the alias name in all small characters.

pkg/worker/overworker.go
```diff
@@ -3,6 +3,7 @@ package worker
 import (
 	"context"
 	"fmt"
+	workerConfig "github.com/giongto35/cloud-game/pkg/worker/config"
```
sadlil: Also, Please run `make fmt` (gofmt, goimports) to fix the formatting issues.

cmd/overworker/main.go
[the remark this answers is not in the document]
giongto35: Yes, I think it's better to have config in root directory: config/worker, so we can have config easily sharable and accessible between package. It also reduces all the dependencies.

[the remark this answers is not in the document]
giongto35: and as sadlil mentioned, in Golang package should be all lower case.

pkg/emulator/libretro/image/scale.go
```diff
@@ -1,7 +1,9 @@
 package image
 
 import (
+	"golang.org/x/image/draw"
 	"image"
+	"image/color"
 )
 
 const (
```
giongto35: Golang has something like 
https://github.com/golang/go/wiki/Iota. I'm ok with the current, just share with you some Go info :P.

pkg/emulator/libretro/image/scale.go
```diff
@@ -11,70 +13,122 @@ const (
...
+
 	// !to do set it once instead switching on each iteration
 	// !to do skip resize if w=vw h=vh
 	switch scaleType {
 	case ScaleSkip:
-		skip(fn, w, h, packedW, vw, vh, bpp, data, image)
+		skip(fn, w, h, packedW, vw, vh, bpp, data, src, out)
 	case ScaleNearestNeighbour:
-		nearest(fn, w, h, packedW, vw, vh, bpp, data, image)
+		draw.NearestNeighbor.Scale(out, out.Bounds(), src, src.Bounds(), draw.Src, nil)
```
giongto35: OMG golang has built-in support @_@. Thanks for finding that out. Sorry if you took time to re-implement it :-s

pkg/emulator/libretro/image/scale.go
```diff
@@ -11,70 +13,122 @@ const (
...
 			index *= bpp
 
-			fn(data, index, j, i, image)
+			out.Set(x, y, fn(data, index))
+		}
+	}
+}
+
+// This implementation has some color bleeding issues
+func bilinear(fn Format, w int, h int, packedW int, vw int, vh int, bpp int, data []byte, _ *image.RGBA, out *image.RGBA) {
```
giongto35: Good work !!!

pkg/worker/config/config.go
giongto35: Yes, I think all config are kept in the same config place.

pkg/worker/overworker.go
sergystepanov: Done.

sergystepanov:
Moved config, hope it's that what you needed.

giongto35:
Merged. Thank you for this awesome work :D!

PR #113, opened 2019-10-17
Change separator in the save file path allowing it to work on Windows

sergystepanov:
This "|" symbol doesn't work with Windows paths, maybe it's better to change it on something cross compatible.

### DIFF
--- a/pkg/worker/room/room.go
+++ b/pkg/worker/room/room.go
@@ -50,6 +50,8 @@ type Room struct {
 	//meta emulator.Meta
 }
 
+const separator = "___"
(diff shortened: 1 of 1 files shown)


giongto35:
Thank you for that change (y). That's separator is indeed not for path. It's just the sign for splitting game name and room hash.
With something like:
29dc1297cd2a621b|Mega Man X4 (U) [SLUS-00561]
I know the game to load is **Mega Man X4 (U) [SLUS-00561]**  and the room hash is **29dc1297cd2a621b**

Issue #114, opened 2019-10-17
[bug] The server returns a bunch of duplicate addresses for the ping check 


sergystepanov:
It seems the server unnecessary duplicates addresses for latencyCheck. Is this a bug (I haven't looked at this part yet)?
As example, two overworkers on localhost:
![image](https://user-images.githubusercontent.com/846874/67036564-87255c00-f124-11e9-8176-631366723a2f.png)
And here is current cloudretro.io:
![image](https://user-images.githubusercontent.com/846874/67036798-eedba700-f124-11e9-9707-42bf42186728.png)

giongto35:
It working for me. 

![image](https://user-images.githubusercontent.com/7755543/67037936-0df02e00-f151-11e9-8a02-b539fc10922c.png)

I just recently updated something on client-side and it may cause malfunction if the webpage is still cached on your side. Can you clear browser cache (ctr shift R) and check again.

sergystepanov:
Check a list from the server, not the list after check what you're showing on the picture.
This:
![image](https://user-images.githubusercontent.com/846874/67039295-b2f71080-f129-11e9-82b1-741eb954ddd4.png)

giongto35:
ah I see, the duplication is because I run few processes (3-5) game instances per host, so I can maximize the CPU.
And yes, it seems redundant to check the same host multiple times. I think you can optimize that part also :D Thank you for that.

giongto35:
If you see anything on client-side part that doesn't make sense, feel free to modify it also. I'm not good at write frontend (:

sergystepanov:
Ok, I'll fix it on the server. To change a list on a map will be enough, I guess.

PR #115, opened 2019-10-17
Don't duplicate addresses for ping check

sergystepanov:


sergystepanov:
To test. Fixes #114.

pkg/overlord/handlers.go
```diff
@@ -274,22 +273,26 @@ func findBestServerFromBrowser(workerClients map[string]*WorkerClient, client *B
...
+		splitter = ","
 	}
 
 	// send this address to user and get back latency
-	log.Println("Send sync", addressList, strings.Join(addressList, ","))
-	data := client.SyncSend(cws.WSPacket{
-		ID:   "checkLatency",
-		Data: strings.Join(addressList, ","),
-	})
+	log.Println("Send sync", addressList)
```
giongto35: Yes, Golang philosophy is to simplify coding :p
Can we create an utitlity function: 
```
// unify will deduplication a list.
func unify(string []string)[]string .
```
So the code can just apply this func directly

so the flow can be just
```
addressList = util.Unify(addressList)
data := client.SyncSend(cws.WSPacket{ID: "checkLatency", Data: strings.Join(addressList, ",")})
```

sergystepanov: I think I'll pass this one.

PR #131, opened 2019-11-20
Unblock sending audio channel

giongto35:
Some deadlock is still happening, causing the app hanging. I think better reuse non-block sending audio first to check if it's really the issue.
I think the audio quality won't get affected much.


sergystepanov:
I think it should not affect the quality at all, I've tried some games with a skip audio frames counter in the non-blocking branch and I've been getting zero drops. 
And what about blocking image channel?
```go
// nanoarch.go:113
NAEmulator.imageChannel <- outputImg
```

giongto35:
@sergystepanov oh I didn't realize you reply here :D .
I think this is because I was trying to find where is the issue causing deadlock. I figured it out yesterday and fixed in this PR.
https://github.com/giongto35/cloud-game/pull/144/files 

The way I'm doing to find deadlock is by running profile. If you access http://localhost:9000/debug/pprof/goroutine?debug=1 you will see where each goroutine waiting on each channel. :D

PR #137, opened 2019-11-23
Update pcsx rearmed core version

sergystepanov:
So, I think I've found the problem with PSX controls. It seems the original version `r22 4b353f8` which you have in the repo had a bug when `start` and `select` buttons won't work. After update to the latest nightly version it'll be working, tested on Ubuntu.


giongto35:
Awesome!! Thank you very much. I also keep wondering since yesterday :p You are a mystery solver.

PR #159, opened 2020-04-08
Add Github release workflow

sergystepanov:
New release workflow for Github CI.

Tested on win and nix.
Hardest part is **** ***** macOS. Not quite sure if it works.
To run: create new release on Github with v{anything} tag or push that tag with git.


giongto35:
Wow. Awesome work. Thank you very much
Now I learn about Github workflow from your diff, it is really nice. With the release, it will be much easier for other people now

PR #161, opened 2020-04-10
Add build workflow

sergystepanov:
Just to add separate multi-os build workflow for the master branch only on every commit. I assume that the master branch is the main release branch.
Actually, this workflow needs run tests `make dev.tools test` but they are broken now.


giongto35:
oooooh this looks so cool @_@. I will merge first and try with the build flow. I think I need to write some test so that can see the effect clearer : p

PR #167, opened 2020-04-20
Refactor input poll

sergystepanov:
Just to refactor slow code.
On my lazy tests it works x2-3 faster (0.4ms+ -> 0.1ms). Will be noticeable on the 1 ms polling interval (the minimum for JS intervals is 4 ms as described in HTML5 specs).

Needs testing.

### DIFF
--- a/web/js/input/input.js
+++ b/web/js/input/input.js
@@ -1,13 +1,9 @@
 const input = (() => {
-    const INPUT_HZ = 100;
-    const INPUT_STATE_PACKET = 1;
-    const KEY_BITS = [KEY.A, KEY.B, KEY.X, KEY.Y, KEY.L, KEY.R, KEY.SELECT, KEY.START, KEY.UP, KEY.DOWN, KEY.LEFT, KEY.RIGHT];
+    let pollIntervalMs = 10;
+    let pollIntervalId = 0;
+    let isStateChanged = false;
 
-    let gameInputTimer = null;
-    let unchangePacket = 0;
-
-    // Game controller state
-    let keyState = {
+    let controllerState = {
         // control
         [KEY.A]: false,
         [KEY.B]: false,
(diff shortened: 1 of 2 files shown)


giongto35:
How did you come with the number :-s ? Can you explain a bit about the reason it will be an improvement? (what is the drawback of old implementation, how it is solved with the new way)

sergystepanov:
First of all, the current polling algorithm with setInterval is quite bad. Due to inherited problems with it (can be blocked and interval is not guaranteed, heap trash). No idea yet what to use instead (requestAnimationFrame, maybe).
Second, this refactor is just for readability and removing bad code.
I compared old and new exec time before pushing into the RTC. With new code the mean exec time was around 0.1ms, old code -- up to 0.5ms. (:

This code is ugly and slow, hope, you understand why. It's noticeable even with 12 elements in array (sic!). There is shouldn't be 2x-3x difference in exec time.
```
// pack keys state
let bits = '';
KEY_BITS.slice().reverse().forEach(elem => {
  bits += keyState[elem] ? 1 : 0;
});
let data = parseInt(bits, 2);
```

giongto35:
Thank you for the very detailed explanation. I got it now. Your refactor on that part is more performant (but I still not sure why 12 elements array is noticeable :-?). As it is late here, I will do a quick check tomorrow to see if it works logicalwise.  :+1:

sergystepanov:
I was thinking about your poll algorithm and found a pretty straightforward bug:
a key press state can be reset only with the corresponding keyup event but you can easily miss it. As example, if you lose browser focus while pressing any keys they won't reset on focus return, only when you press again the same keys.

(on the gif it hard to tell because of encoding artefacts, but I'm holding left key before alt-tabbing from the window and going back where it still active)
annoy03

giongto35:
I'm aware of this key stuck issue when playing. It is indeed troublesome. 
I think on backend we can keep the last state update time and reset everything when the last state update time is > 1second.

giongto35:
Btw, I'm testing your change now :D

PR #168, opened 2020-04-20
Add generic application statistics overlay

sergystepanov:
This feature adds an option to gather and display various statistics on the client side.
The statistics module includes some examples of the submodule structure in order to be able to write your own client or server params tracking code.
By default the feature toggle is bound to the `backspace` key.
Display module of the feature uses and depends on the RAF callbacks for smoother animation limited by default to 32fps.

feature003

PR #169, opened 2020-04-22
Fix ui scaling for Chrome

sergystepanov:
Because of the fact that "gameboy" interface has absolute dimensions and scaled with the browser window, everything inside it (DOM elements) scales as well. The scaled size can have odd dimensions which leads Chrome to render everything blurred. As example a picture below (before/after).
This fix helps to mitigate the issue to some extent: and just forces the target dimensions have nearest even values.

image

Needed for #168.

### DIFF
--- a/web/js/env.js
+++ b/web/js/env.js
@@ -1,14 +1,15 @@
 const env = (() => {
-    const win = $(window);
+    // UI
     const doc = $(document);
+    const gameBoy = $('#gamebody');
+    const ghRibbon = $('#ribbon');
 
-    // Screen state
     let isLayoutSwitched = false;
 
     // Window rerender / rotate screen if needed
     const fixScreenLayout = () => {
-        let targetWidth = doc.width() * 0.9;
-        let targetHeight = doc.height() * 0.9;
+        let targetWidth = Math.round(doc.width() * 0.9 / 2) * 2,
+            targetHeight = Math.round(doc.height() * 0.9 / 2) * 2;
 
         // mobile == full screen
         if (env.getOs() === 'android') {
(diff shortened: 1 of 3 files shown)


web/js/env.js
```diff
@@ -1,14 +1,15 @@
 const env = (() => {
-    const win = $(window);
+    // UI
     const doc = $(document);
+    const gameBoy = $('#gamebody');
```
giongto35, line 4: it's not gameBoy though, can name it gamebody as well :D

giongto35:
this is sophisticated. Don't know that even dimension and odd dimension have this significant difference on rendering. Thank you very much.

sergystepanov:
> this is sophisticated. Don't know that even dimension and odd dimension have this significant difference on rendering. Thank you very much.

Odd or even doesn't matter, what matters is just if you don't scale by a round number (or even a even (: number).  But the main culprit is `transform: translate(-50%, -50%)`, I'm not sure why it leads to a blurry rendering in Chrome, found some posts about this bug, but even dimensions did the trick for me. (my guess, its rendering engine interpolate the whole picture shifted into the space between the pixels)

web/js/env.js
sergystepanov: I renamed it for a better clarity. What does a `gamebody` even mean? `Gameboy` at least something relatable and can be associated with the part of the interface.

sergystepanov: Oh, forgot that you have to have ™ mark after it (: GameBoy™

giongto35:
Intersting issue :D So translate(0.5,0.5) will make the image shift to (1, 1) with blur. If we translate(1,1) directly, it will not blur right? 
I also want to give a try. How did you test the issue? did you just scale in and out?

sergystepanov:
I'm not sure how exactly this rendering stuff works, I need to read about it. As example here is one SO [question](https://stackoverflow.com/questions/14677490/blurry-text-after-using-css-transform-scale-in-chrome). If I understand correctly, with use of absolute values for matrix transformations (css translate) instead of relative values it should be rendered sharp because we won't shift anything between pixels. It's just how bilinear filtering works: you translate a pixel onto some fraction position and it have to interpolate surroundings.
I tested on W10 and latest Chrome.
Right now, for me, after I open the app, while it checking the best server everything is sharp but after the moment it render the menu it goes to blur (sic!). Next, when I scale the window most of the times it's blurred. With my fix it stays sharp, just maybe 5% of blur is left.

PR #175, opened 2020-04-27
Add support firefox safari ios

trichimtrich:
I did some test with webrtc of browser api and pion. [Ref Repo](https://github.com/trichimtrich/testrtc2)
The current approach is (assume browser <-> worker can see each others)
- Browser creates a peer instance
- Browser createDataChannel with `ordered: true, negotiated: true, id: 0`
- Browser createOffer and sends its SDP to worker via `initwebrtc`
- Worker creates a peer instance
- Worker also createDataChannel with `ordered: false, negotiated: true, id: 0`
- Worker adds 2 tracks
- Worker set remote SDP (from browser)
- Worker createAnswer and sends its SDP to browser via `initwebrtc` (reply)

It acts well on Chrome (Desktop + Android), but not on Safari + Firefox (Desktop + Mobile) this way.
Edge 44 is out of this testing scope because it doesn't support `addTransceiver` and `createDataChannel`, Edge Chromium is considered as Chrome.

I changed the process to normal
- Browser asks for rtc session, sends `initwebrtc`
- Worker adds data channel `ordered: true, negotiated: false`
- Worker adds 2 tracks
- Worker createOffer and sends SDP
- Browser adds remote SDP from worker
- Browser createAnswer and sends SDP
- Worker adds remote SDP from browser
- During that they both exchange ICE candidate.

It works on Firefox + Safari / iOS now
image

I also merged the `browserClient` and `session` structs of coordinator, and cleaned up a bit in backend.


giongto35:
<3 <3 <3 This is gold. It helps solve all the headaches I got.

giongto35:
Can you explain the difference between two approach. Why sending offer from server works but from client not.

trichimtrich:
The different is the stream provider (tracks + datachannel) is the one makes offer.

Just did by the blackbox method, not sure how each browsers implement and use webrtc internally, but I tried the current approach (before fix).
- Chrome worked well
- Firefox received OnTrack event, but seems couldn't receive the stream (or maybe not rendered).
- Safari did not trigger OnTrack event

So I suppose they use a normal way, where the provider knows what properties of its data (codec, connection type, ...) to make offer.

pkg/worker/coordinator.go
```diff
@@ -118,13 +183,16 @@ func (h *Handler) RouteCoordinator() {
...
 			log.Println("Received a quit request from coordinator")
-			session, ok := h.sessions[resp.SessionID]
-			log.Println("Find ", resp.SessionID, session, ok)
+			session := h.getSession(resp.SessionID)
 
-			room := h.getRoom(session.RoomID)
-			// Defensive coding, check if the peerconnection is in room
-			if room.IsPCInRoom(session.peerconnection) {
-				h.detachPeerConn(session.peerconnection)
+			if session != nil {
```
giongto35, line 188: golang good practice is
if session, ok := h.sessions[resp.SessionID]; ok {
   // do something
} else {
   //
}

use ok to check

pkg/worker/coordinator.go
```diff
@@ -47,48 +49,110 @@ func (h *Handler) RouteCoordinator() {
...
 			if err != nil {
-				panic(err)
+				log.Println("Error: Cannot decode json:", err)
+				return cws.EmptyPacket
 			}
-			localSession, err := peerconnection.StartClient(initPacket.SDP, initPacket.IsMobile, iceCandidates[resp.SessionID])
+
+			localSession, err := peerconnection.StartClient(
+				initPacket.IsMobile,
+				func(candidate string) {
```
giongto35, line 71: well implemented

giongto35:
Very good job @trichimtrich !
So I have a question:
- If server doesn't send back ICE candidate, do you think it still works? I think our server is fully open, so one candidate is enough?

pkg/coordinator/handlers.go
```diff
@@ -80,151 +84,186 @@ func (o *Server) getPingServer(zone string) string {
...
+		workerID = uuid.Must(uuid.NewV4()).String()
+		// check duplicate
+		if _, ok := o.workerClients[workerID]; !ok {
+			break
+		}
+	}
+
+	// Create a workerClient instance
+	wc := NewWorkerClient(c, workerID)
+	wc.Println("Generated worker ID")
```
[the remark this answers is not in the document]
trichimtrich: Yes, the formal way looks more clean. The structure log makes it easier to track which context and data we working on. Also, we should maintain an exported file too.

pkg/coordinator/handlers.go
```diff
@@ -80,151 +84,186 @@ func (o *Server) getPingServer(zone string) string {
...
-			ID:   "serverID",
-			Data: serverID,
-		},
-		nil,
-	)
-	o.RouteWorker(client)
-
-	client.Listen()
+
+	// Create a workerClient instance
```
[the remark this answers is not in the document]
trichimtrich: The reason of this change is because the log handler. If we can use structure log, the change can be discard. Totally agree

pkg/worker/coordinator.go
trichimtrich: the `h.getSession(resp.SessionID)` uses the same mechanism, not sure why `getSession` func was there before

trichimtrich:
Thanks ; )
Yes for the ICE candidate, 1 side is enough in most of my test. I did mostly on local network, so basically they should easily connect to each other.

But also the ICE candidates are infos of one peer (from STUN) need to let other peer know.
In this case our server is in public interface, so server ICEs need to be sent to browser.

But I suggest to exchange between both party, the cost is small, but guarantee that we do it right.

Issue #158, opened 2020-04-03
Rotated image on some games (mainly flight simulators) - FBNeo


jk7gr:
Dear all,
I am amazed by your work. I have spend a lot of time to understand and build my server. I understand parameterization of the core is very limited at the moment with dimensions and aspect ration to be provided. However the issue I am facing is that some games run rotated by 90 decrease to maintain a landscape type of ratio.
![image](https://user-images.githubusercontent.com/17036347/78334310-36030a80-7594-11ea-8de4-64c3f922dd0a.png)
Is this something I can configure?
Thank you!

sergystepanov:
Hi @jk7gr! It is always good to hear that more people find a common interest in this project. (:
Speaking about rotated game video, I've never thought about this problem until now. 
At the moment it's not possible to a user rotate the screen (actually, you can turn your monitor or smartphone with your own hands (: ) but manually you can do something like this:

![image](https://user-images.githubusercontent.com/846874/78348792-48d50980-75ab-11ea-93d6-209974752ab5.png)

I think, in order to add this feature we should first discuss the best way of its implementation from a user perspective. To add client-side options?
From the dev perspective, I think, the best and easiest way is to rotate canvas in a browser like on the picture above. Guess, It won't be CPU (or GPU, because right now rendering stuff in browsers is GPU accelerated) demanding, by any stretch.
I will try to look how it's implemented in Retroarch.

jk7gr:
hi @sergystepanov really much appreciate the fast response here. 
I think the existence of a rotate button is fine for the moment. It will allow the run of the content because currently is hard to rotate the TFT panel :) (smartphone I agree the rotation is easier). 
In the long run, if this is somehow tackled behind scenes, it will allow a cleaner GUI.

sergystepanov:
I believe we can make rotation work automatically based on core-provided w/h to aspect ratio mismatch: as example a game on the screenshot has 320/240 [1.(3)] != 240/320 [0.75] then call Libretro core rotation with 90* CCW but I not sure if all the cores rotate portrait games that way.

https://github.com/libretro/RetroArch/blob/1d7cae2ceb6c6dcb00e0b2b43dd49df2d19bda22/libretro-common/include/libretro.h#L489-L494

![image](https://user-images.githubusercontent.com/846874/78365881-62d11500-75c8-11ea-9d80-e2168afbb74e.png)

jk7gr:
dear all,
any news on this?

Thank you!

[...]

PR #186, opened 2020-05-23
Add initial support of portrait games

sergystepanov:
See #158.

Needs testing & code review.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"flag"
+	"github.com/giongto35/cloud-game/pkg/emulator/libretro/image"
 	"time"
 )
 
@@ -50,6 +51,7 @@ type EmulatorMeta struct {
 	BaseWidth       int
 	BaseHeight      int
 	Ratio           float64
+	Rotation        image.Rotate
 }
 
 var EmulatorConfig = map[string]EmulatorMeta{
--- a/pkg/emulator/libretro/image/color.go
+++ b/pkg/emulator/libretro/image/color.go
@@ -15,7 +15,7 @@ const (
 
 type Format func(data []byte, index int) color.RGBA
 
-func rgb565(data []byte, index int) color.RGBA {
+func Rgb565(data []byte, index int) color.RGBA {
 	pixel := (int)(data[index]) + ((int)(data[index+1]) << 8)
 
 	return color.RGBA{
@@ -26,7 +26,7 @@ func rgb565(data []byte, index int) color.RGBA {
 	}
 }
 
-func rgba8888(data []byte, index int) color.RGBA {
+func Rgba8888(data []byte, index int) color.RGBA {
 	return color.RGBA{
 		R: data[index+2],
 		G: data[index+1],
--- a/pkg/emulator/libretro/image/draw.go
+++ b/pkg/emulator/libretro/image/draw.go
@@ -4,15 +4,57 @@ import (
 	"image"
 )
 
-func DrawRgbaImage(pixFormat int, scaleType int, w int, h int, packedW int, vw int, vh int, bpp int, data []byte, image *image.RGBA) {
-	switch pixFormat {
-	case BIT_FORMAT_SHORT_5_6_5:
-		Resize(scaleType, rgb565, w, h, packedW, vw, vh, bpp, data, image)
-	case BIT_FORMAT_INT_8_8_8_8_REV:
-		Resize(scaleType, rgba8888, w, h, packedW, vw, vh, bpp, data, image)
-	case BIT_FORMAT_SHORT_5_5_5_1:
-		fallthrough
-	default:
-		image = nil
+type imageCache struct {
+	image *image.RGBA
+	w int
+	h int
+}
+
+var canvas = imageCache{
+	image.NewRGBA(image.Rectangle{}),
+	0,
+	0,
+}
+
+func DrawRgbaImage(pixFormat Format, rotationFn Rotate, scaleType, w, h, packedW, bpp int, data []byte, dest *image.RGBA) {
+	if pixFormat == nil {
+		dest = nil
+	}
+
+	// !to implement own image interfaces img.Pix = bytes[]
+	ww, hh := w, h
+	if rotationFn.IsEven {
+		ww, hh = hh, ww
+	}
+	src := getCanvas(ww, hh)
+
+	drawImage(pixFormat, w, h, packedW, bpp, rotationFn, data, src)
+	Resize(scaleType, src, dest)
+}
+
+func drawImage(toRGBA Format, w, h, packedW, bpp int, rotationFn Rotate, data []byte, image *image.RGBA) {
+	for y := 0; y < h; y++ {
+		for x := 0; x < w; x++ {
+			src := toRGBA(data, (x+y*packedW)*bpp)
+			dx, dy := rotationFn.Call(x, y, w, h)
+			i := dx*4 + dy*image.Stride
+			dst := image.Pix[i : i+4 : i+4]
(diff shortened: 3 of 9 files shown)


sergystepanov:
Funny thing that a call to the map `Angles[angle]` in a loop was a huge performance hit (:
Do you know why is that? Go maps are slow or something?

giongto35:
@sergystepanov do you mean : pkg/emulator/libretro/image/rotation_test.go :179? How did you figure out the performance was hit because of that?

sergystepanov:
I already fixed it.
I didn't expect that it will be so much expensive so the games started lagging (: a lot.
I mean that there was a call to the map inside [this double loop](https://github.com/giongto35/cloud-game/pull/186/files#diff-ab1eafce627bacd812e4f1b840d3e984R35-R48).

sergystepanov:
Ah, I got what you mean (: Right, I think that's the problem. Thanks.
a := b

And for each assignment it did:
> That is, the assignment “f := MyFunc” would generate code like:
> MOV $MyFunc·f(SB), f1
> DATA MyFunc·f(SB)/8, $MyFunc(SB)
> GLOBL MyFunc·f(SB), 10, $8

In the current code it's being assigned only once per game start.

sergystepanov:
```
goos: windows
goarch: amd64
pkg: github.com/giongto35/cloud-game/pkg/emulator/libretro/image
BenchmarkDirect
BenchmarkDirect-4      	1000000000	         0.916 ns/op
BenchmarkAssign
BenchmarkAssign-4      	305344443	         4.57 ns/op
BenchmarkMapDirect
BenchmarkMapDirect-4   	42606237	        27.0 ns/op
PASS
```

```go
func BenchmarkDirect(b *testing.B) {
	for i := 0; i < b.N; i++ {
		p1, p2 := Rotate90(1, 1, 2, 2)
		p1, p2 = p2, p1
	}
}


func BenchmarkAssign(b *testing.B) {
	fn := Angles[Angle90].Call
	for i := 0; i < b.N; i++ {
		p1, p2 := fn(1, 1, 2, 2)
		p1, p2 = p2, p1
	}
}

func BenchmarkMapDirect(b *testing.B) {
	for i := 0; i < b.N; i++ {
		p1, p2 := Angles[Angle90].Call(1, 1, 2, 2)
		p1, p2 = p2, p1
	}
}
```

sergystepanov:
What the actual f***:
```
goos: windows
goarch: amd64
pkg: github.com/giongto35/cloud-game/pkg/emulator/libretro/image
BenchmarkCustomType
BenchmarkCustomType-4      	38711425	        26.2 ns/op
BenchmarkPrimitiveType
BenchmarkPrimitiveType-4   	136776249	         8.58 ns/op
PASS
```

```go
func BenchmarkCustomType(b *testing.B) {
	fns := map[Angle]func(x, y, w, h int) (int, int){
		Angle90: Rotate90,
	}

	for i := 0; i < b.N; i++ {
		p1, p2 := fns[Angle90](1, 1, 2, 2)
		p1, p2 = p2, p1
	}
}

func BenchmarkPrimitiveType(b *testing.B) {
	fns := map[uint]func(x, y, w, h int) (int, int){
		0: Rotate90,
	}

	for i := 0; i < b.N; i++ {
		p1, p2 := fns[0](1, 1, 2, 2)
		p1, p2 = p2, p1
	}
}
```

giongto35:
OMG, this is so tricky. Thanks for the investigation.
Why the itoa affect the performance so much:/

sergystepanov:
Oh, wow. This' because of uint8 key. Nice language...really like it.
```
goos: windows
goarch: amd64
pkg: github.com/giongto35/cloud-game/pkg/emulator/libretro/image
BenchmarkUintKey
BenchmarkUintKey-4    	129952644	         8.73 ns/op
BenchmarkUint8Key
BenchmarkUint8Key-4   	47733645	        25.8 ns/op
BenchmarkIntKey
BenchmarkIntKey-4     	139873735	         8.73 ns/op
PASS
```

```go
func BenchmarkUintKey(b *testing.B) {
	fns := map[uint]func(x, y, w, h int) (int, int){
		0: Rotate90,
	}

	for i := 0; i < b.N; i++ {
		p1, p2 := fns[0](1, 1, 2, 2)
		p1, p2 = p2, p1
	}
}

func BenchmarkUint8Key(b *testing.B) {
	fns := map[uint8]func(x, y, w, h int) (int, int){
		1: Rotate90,
	}

	for i := 0; i < b.N; i++ {
		p1, p2 := fns[1](1, 1, 2, 2)
		p1, p2 = p2, p1
	}
}

func BenchmarkIntKey(b *testing.B) {
	fns := map[int]func(x, y, w, h int) (int, int){
		1: Rotate90,
	}

	for i := 0; i < b.N; i++ {
		p1, p2 := fns[1](1, 1, 2, 2)
		p1, p2 = p2, p1
	}
}
```

sergystepanov:
Added benchmarks. With uint key it's x3 faster but still x3 slower (: because of the map for some reason.

sergystepanov:
Fixed with arrays (:/

giongto35:
Im just wondering how should we interpret the benchmark. I think at ns level, these optimization like int -> uint can show significant impact but it might not be the real bottleneck.
For example, if an operation takes 24ns, we remove one line that cost 16ns, it reduces to 6ns and we think it is faster by 3 times, but in fact it just faster by 16 ns.
In the real run, the loop is like 60 fps, I think the game getting lagger overtime is due to some sort of memory piled up. 
Anyway, I will checkout and investigate the issue as well. : D Thank you so much for bringing this issue up.

sergystepanov:
I'm not exactly sure what's your point (: with 16 ns or 3x faster, and you mean 8 ns? This op will be faster both: 3x and by 16 ns. If we'll use just this one operation one time as it's right now, that's doesn't matter at all. If we **have to** use this op inside a O(N) algo (as I did initially) or worse (because some algos can't have time complexity lower some upper bound even in a best case Ω) then there's no choice but either change your algo complexity and if you can't -- optimize operations inside.
Doing this bench, you can imagine how it would impact your performance isolated from everything else: 16ns x N. And in my case it was 320x240x1x32ns (1.5ms) then times it 60 fps.

In the real world you just look at ELK graphs of some main profiled code blocks or more likely just response times of your users and CPU/MEM utilization. If it's a trash you're profiling this on staging environment, then test fix, then ship to prod.

Now come to think about it: we have two separate data flows, one for video processing and one for audio processing. Audio processing is much-much faster than video. And I believe we don't have any king of AV synchronization. In that case if video takes more time for processing it's matter of time when it goes out of sync. This' exactly what I'm getting -- audio is being severe ahead of video especially when the OS scheduler pauses goroutines (app threads). Optimizing this video processing time should help but sadly -- not much. This issue is kinda critical, actually. (: But it needs some research. I'm lazy. (:

sergystepanov:
Let's focus on the issue with portrait games (:
People are waiting. Performance wise it's alright -- rotation takes < 1% of time of the draw function.
Just check the code overal and if you have time -- test some portrait and your selection of ROMs for video artefacts. That would be great.

giongto35:
I just ran your code bench and saw all the metrics look fine, then I realized it was after your change from array to map. Map is usually a bottleneck and it was an example in Go profile : https://blog.golang.org/pprof.
Overall, the code looks good to me : D Thank you very much for taking this change + nice investigation.

sergystepanov:
Cool, thanks. Will read the docs. Next I'll finish settings module and then, maybe, try to understand [RTP Sync](https://github.com/runner365/read_book/blob/master/RTP_RTCP/RTP%EF%BC%9AAudio%20and%20video%20for%20the%20Internet.pdf). %)

Issue #189, opened 2020-06-01
Can't find the game when reloading the game from browser


akh64bit:
I am running the game locally on chrome.
When I go to localhost:8000 for the first time after launching the docker, the game launches successfully. Here is the logs from the console. 
[input] keyboard has been initialized
log.js?v=3:9 [input] joystick has been initialized
log.js?v=3:9 [input] touch input has been initialized
log.js?v=3:9 [ws] <- open connection
log.js?v=3:9 [ws] -> setting ping interval to 2000ms
log.js?v=3:9 [ping] <-> {http://localhost:9000/echo: 5}
log.js?v=3:9 [rtcp] <- received stunturn from worker [{"urls":"stun:stun.l.google.com:19302"},{"urls":"stun:127.0.0.1:3478"},{"urls":"turn:127.0.0.1:3478","username":"root","credential":"root"}]
log.js?v=3:9 Added voice track
log.js?v=3:9 [rtcp] ice gathering
log.js?v=3:9 [rtcp] got ice candidate: {"candidate":"candidate:3792981046 1 udp 2122260223 10.0.0.26 53547 typ host generation 0 ufrag 85jv network-id 1 network-cost 10","sdpMid":"0","sdpMLineIndex":0}
log.js?v=3:9 [rtcp] ice gathering completed
log.js?v=3:9 [rtcp] <- iceConnectionState: connected
log.js?v=3:9 [rtcp] connected...
log.js?v=3:9 [rtcp] <- iceConnectionState: disconnected
log.js?v=3:9 [rtcp] disconnected...

However when I reload the game from browser, I can't see any game in the game window. Here are the logs when it fails to reload.
[input] keyboard has been initialized
log.js?v=3:9 [input] joystick has been initialized
log.js?v=3:9 [input] touch input has been initialized
log.js?v=3:9 [ws] <- open connection
log.js?v=3:9 [ws] -> setting ping interval to 2000ms

I have to run the docker again for launching the game again.
Does anyone know of the issue.

Thanks in advance for your help.

sergystepanov:
Hi @akh64bit,

Could you please show your Docker logs with coordinator / worker output during this lunch / relunch?

giongto35:
@akh64bit 
It sometimes happens because the connection cleaning process is not properly implemented (code bug issue). The worker is panic due to "sending to a closed channel". The bug can be reproduced by keep reloading. I didn't have time to fully fix that.
[...]

PR #191, opened 2020-06-05
Fix high CPU due to nonstop loop

giongto35:
I recently had an implementation to broadcast voice chat. However, I badly wrote it with a constant loop and it caused full CPU usage. This issue also affected docker run because the CPU usage in Docker is capped, making suspicious freeze in the middle of the run. https://github.com/giongto35/cloud-game/issues/189.

### DIFF
--- a/pkg/webrtc/webrtc.go
+++ b/pkg/webrtc/webrtc.go
@@ -198,21 +198,22 @@ func (w *WebRTC) StartClient(isMobile bool, iceCB OnIceCallback) (string, error)
 	})
 
 	w.connection.OnTrack(func(remoteTrack *webrtc.Track, receiver *webrtc.RTPReceiver) {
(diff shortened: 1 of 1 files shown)


pkg/webrtc/webrtc.go
```diff
@@ -198,21 +198,22 @@ func (w *WebRTC) StartClient(isMobile bool, iceCB OnIceCallback) (string, error)
...
-		rtpBuf := make([]byte, 1400)
-
-		log.Println("Received Voice from Client")
-		for {
-			if w.RoomID == "" {
-				// skip sending voice when game is not running
-				continue
-			}
-
-			i, err := remoteTrack.Read(rtpBuf)
```
giongto35, line 210: Even if I move RoomID check after, it still caused high CPU. Need to check if there is buffer mechanism in remoteTrack

PR #195, opened 2020-06-10
N64 Support

88hcsif:
Add support for OpenGL cores in general but only tested with Mupen64plus and Mupen64plusNext on a Raspberry Pi.
It probably needs more work to run on every system and with other OpenGL libretro libraries.
Also adds support for analog stick from frontend to backend.
Also adds CLI settings that allow to run coordinator and worker over HTTPS on different setups.


giongto35:
This is a lot of change, so I think I will take time reviewing it. Hope I can finish it by this weekend.
Thank you so much for your effort. 
*btw, I don't have raspberry pi. Is there any alternative way to test? Do you have any screenshot, I'm really curious about the result?

88hcsif:
Yeah, no worries, it is totally understandable that it will take time.

I actually managed to run it on a x86_64 debian based linux distribution and it worked fine.
With the last commit I added the mupen64plus core for x86_64 so I'd be keen to know if it works for you without the Raspberry Pi.
The only caveat is that the distribution cannot be headless, it must have a graphical interface for OpenGL to work even if nothing is rendered on screen (a dummy window of size 1x1 is still created).
So the worker needs to either be launched in a terminal/launcher from the OS UI or from an ssh session I normally run <code>DISPLAY=:0 ./bin/worker ...</code>.

Alternatively, I can also provide the public hostname of my Pi for anyone to try but I would rather not post it here. Also I do not keep the service constantly running and it can occasionally crash, so we would have to coordinate a test session.

Finally, here is a screenshot of Mario Kart 64:
image
(the thing at the top is a toggle to allow analog stick mode which also changes the appearance of the dpad/stick)

88hcsif:
Also apologies if some of the code is not very well commented.
For example, I forgot to document the reason why the code to call into libretro has got more complicated.
Here is a bit of context on that: https://github.com/golang/go/issues/39411

giongto35:
Thank you for the detailed explaination.
Regarding the , I'm also finding way to run any program on headless server and capture the screen. Don't know if it's possible with X11.

sergystepanov:
@giongto35, could you check on your systems if it works?
For me it does not:
- Windows (Msys2) -- old emulators are ok, n64 crashed (no idea, possible because of i7 2640 OpenGL stuff)
- Windows / Ubuntu 20 (WSL2) -- tried n64 -- crash on x11 (needs special config)
- Ubuntu 19 -- encoder crash on any core:
```
I0613 19:51:45.466531    7840 media.go:138] Video Encoder:  VP8
error create new encoder Failed to initialize encoder
I0613 19:51:45.466647    7840 media.go:73] Enter fan audio
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0xbb2dd2]

goroutine 49 [running]:
github.com/giongto35/cloud-game/pkg/worker/room.(*Room).startVideo.func1(0x0, 0x0)
	/home/sergu/cloud-game/pkg/worker/room/media.go:146 +0x22
github.com/giongto35/cloud-game/pkg/worker/room.(*Room).startVideo(0xc0000d66e0, 0xf0, 0xa0, 0xeff516, 0x3)
	/home/sergu/cloud-game/pkg/worker/room/media.go:151 +0x40e
created by github.com/giongto35/cloud-game/pkg/worker/room.NewRoom.func1
	/home/sergu/cloud-game/pkg/worker/room/room.go:146 +0x7a4
make: *** [Makefile:77: dev.run] Error 2
```

giongto35:
@sergystepanov  haven't got time looking a this or now. My ubuntu is 18.04 and it works fine on other core. I haven't tested with N64 core (need to find a n64 game)

Did you use VM to test on different OS? it is very careful testing indeed : D

sergystepanov:
[There are](https://n64squid.com/homebrew/) some homebrews you might try.
I wouldn't say it's 'a very careful' testing, rather an attempt to run it somehow (:
If I had time, I'd try to run it on a bunch of own rPis / Odroids.

Retroarch won't run for me any of n64 games on any N64 core as well, btw. Either OpenGL driver or hardware issue (but tried to port own 3d stuff back from school some time ago and was able to run it only with GLFW or raw WinAPI with pretty old and painful compatibility profiles). So I can't help here.

If it works OK for you, then it will work for most of people, I guess.

giongto35:
oh, I checked out wrong branch.
Just test the feature/n64. I got
```
# pkg-config --cflags  -- sdl2
Package sdl2 was not found in the pkg-config search path.
Perhaps you should add the directory containing `sdl2.pc'
to the PKG_CONFIG_PATH environment variable
No package 'sdl2' found
pkg-config: exit status 1
```

sergystepanov:
It's [obvious](https://wiki.libsdl.org/Installation#Linux.2FUnix):
```
sudo apt-get install libsdl2-dev
```

giongto35:
@88hcsif : Your commit is well documented so I followed quite easy :D 
I'm curious about the OpenGL implementation. Did you write by yourself or you referred from https://github.com/libretro/ludo?
The code looks good to me. I tried with N64 game and it can launch the game.
The only thing is arrow input doesn't work and can you resolve the conflicting file so I can merge it.

giongto35:
I see you introduced new control pad :/ I think the original one is also doing similar movement right?

giongto35:
@sergystepanov the build is failing, should we merge and fix it later?

sergystepanov:
@giongto35, sure, I'll fix it.

PR #196, opened 2020-06-12
Make it choose ws/wss protocol based on URL's protocol.

sergystepanov:
When you open the app with https: protocol it will chose wss secured socket connection instead of  hardcoded addresses / domains.


sergystepanov:
Check it on your staging or prod first.

PR #197, opened 2020-06-14
Add Github Docker repo publisher

sergystepanov:
Added gh Docker repo publish. Build from Dockerfile.
Should look like [this](https://github.com/sergystepanov/cloud-game/packages/261228).
You can pull them as described there and add own custom stuff.

When you push either to master or tag anything with v-prefixed tag, it will build and publish new Docker image. An image with `latest` tag will be replaced by new master source, v-something will be a build from v-prefixed source.

Sha-tagged images can be used as well but I don't think they are useful, so I won't be using them.


giongto35:
Is this recommended to use this image from master for production? I currently push to http://docker.io/ registry, I can consider pulling from git right?

PR #204, opened 2020-06-20
Add basic N64 support on Mac

88hcsif:
Note: mupen64plus lib unload is not working properly on Mac, so stopping and restarting N64 emulation will cause a crash


88hcsif:
I am not sure why the CI tests are failing now...
If you point me in the right direction I can try to fix it.

sergystepanov:
There is no coordinator package now, probably you've removed it by mistake

88hcsif:
Ah, yeah, I accidentally deleted a file...

88hcsif:
Okay, my branch is really screwed up I will just restart from scratch...

PR #205, opened 2020-06-20
Add basic N64 support on Mac

88hcsif:
Note: mupen64plus lib unload is not working properly on Mac, so stopping and restarting N64 emulation will cause a crash

### DIFF
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -7,15 +7,15 @@ import (
 	"os/signal"
 	"time"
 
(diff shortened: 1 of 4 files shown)


giongto35:
Hi 88hcsif,

Thank you very much for your contribution. I won't be available for the next few days, so it might take time to review your pull requests. Sorry for that :(

88hcsif:
No worries and thanks for the heads up!
I might send some more PRs over if I make some progress but I will keep things separate.

PR #206, opened 2020-06-20
Add frame delay or frame rate stats to stat view.

88hcsif:
Unfortunately Firefox does not expose frame delay stat.
Unfortunately Chrome and Safari do not expose frame rate stat.


giongto35:
Wow, I like this so much. Let me try it on prod to test the performance. Thank you very much for taking this

PR #210, opened 2020-06-26
Free save state buffer from malloc

sergystepanov:
While working on save states, found that we are using one hugely leaking `malloc` function for save state buffering.

https://github.com/giongto35/cloud-game/blob/c23347db0814856ae07e5649886d3d340305d08a/pkg/emulator/libretro/nanoarch/nanoarch.go#L656-L665

The later call of `C.GoBytes` makes a copy of `data`.
Here some test before and after on PSX with 4Mb save states.
Made the PR, because I think, this fix is kinda critical.
Also there are some unidentified leaks still present.

Mashing the save. Memory `never` recovers.

before

Mashing the save. Memory recovers.

before2

Forgot to mention that Libretro cores can change save state size (shrink) dynamically. It's not an issue atm, but if we ever change the current code to fixed max save state buffer length calculated only on the game load, then we may have some problems with malloc's garbage in the buffer. Retroarch uses zeroed calloc buffers for that reason. Just keep it in mind. (:

### DIFF
--- a/pkg/emulator/libretro/nanoarch/nanoarch.go
+++ b/pkg/emulator/libretro/nanoarch/nanoarch.go
@@ -653,13 +653,16 @@ func serializeSize() uint {
 	return uint(C.bridge_retro_serialize_size(retroSerializeSize))
 }
 
-// serialize serializes internal state and returns the state as a byte slice.
+// Serializes internal state and returns the state as a byte slice.
 func serialize(size uint) ([]byte, error) {
 	data := C.malloc(C.size_t(size))
+	defer C.free(data)
+
 	ok := bool(C.bridge_retro_serialize(retroSerialize, data, C.size_t(size)))
 	if !ok {
 		return nil, errors.New("retro_serialize failed")
 	}
+
 	bytes := C.GoBytes(data, C.int(size))
 	return bytes, nil
 }


giongto35:
@sergystepanov  Thanks for pointing this out. 
May I know where is the graph come from? Is it the prometheus metric?
Regarding to the dynamic save size, I derive this code from ludo. I think each time it refetch the serialize size, so maynot be the case

> 
>   | func serializeSize() uint {
>   | return uint(C.bridge_retro_serialize_size(retroSerializeSize))
>   | }

sergystepanov:
@giongto35, it's just a system app for Windows -- [ProcessHacker](https://github.com/processhacker/processhacker) and process info. Nice for simple debugging.

That serialization size according to the API should not grow bigger between core_load_game calls and the normal way to handle this stuff is like in [Retroarch](https://github.com/libretro/RetroArch/blob/966348327db14e7f2c5e5d12c630b2f8c25a8bc7/tasks/task_save.c#L588-L594).
Not sure what way is better and faster. Malloc(getCoreSerializeSize) should be faster.

giongto35:
@sergystepanov oh, I use Ubuntu. That's nice tool. Thanks for that (y)

Issue #198, opened 2020-06-14
Improve WebRTC lag


88hcsif:
So far I have only tried to play Mario Kart 64 with the added N64 support.
I tried battle games with friends and it is playable.
But as soon as I try to do a race even with one player the lag becomes too long.

I checked and the Raspberry Pi is coping fine with the emulation and the video encoding.
Mario Kart 64 runs at around 29 frames per second.
Between receiving an input change from the client, giving it to the emulator, drawing the next frame, encoding in h264 or vp8, and sending the frame takes anything between 50 and 100 ms.

I think the problem is in the video player on the client side (this is on Chrome).
Probably, I will make a PR at some point to add this extra data to the stats view.
If I get the stats for the RTCPeerConnection and I look at the video track while playing a race I can see:
<pre>
Report: track
ID: RTCMediaStreamTrack_receiver_52
Timestamp: 1592167875098.266
trackIdentifier: video
remoteSource: true
ended: false
detached: false
kind: video
jitterBufferDelay: 77.946
jitterBufferEmittedCount: 2883
frameWidth: 320
frameHeight: 240
framesReceived: 2945
framesDecoded: 2883
framesDropped: 0
</pre>

There is a big difference between framesDecoded and framesReceived.
At 29 fps the lag due to this is about 2 seconds.
I am completely new to WebRTC so I am not sure exactly what is going on...

sergystepanov:
WebRTC is _amazing_ piece of technology (: It's been 10 years and it is still in drafting experimental state even today, after 2 years since a first stable release, with a lot of bugs in their tracker.
I thought before about this lip sync (how it's called in WebRTC) issue but I haven't had time to try to understand it and read some [books](https://github.com/runner365/read_book/blob/master/RTP_RTCP/RTP%EF%BC%9AAudio%20and%20video%20for%20the%20Internet.pdf). What I can tell for sure: it is not Libretro issue. My guess, since we are not using any kind of A/V synchronization when encoding two separate streams in two separate (sic!) threads (channels), it's only matter of time when everything goes out of sync.
If you want to reproduce this issue, just make the OS scheduler to starve worker thread (with heavy load on the same system or something like that). After that you will get a couple second delay of the video stream. It is strange that A/V delays are going down on the graphs but WebRTC mixer will be playing it with a constant desync until session reset.
[...]

PR #211, opened 2020-06-26
Explicitly set RTP timestamp to cope with variable frame rate.

88hcsif:
Set the timestamp as early as possible and propagate it through the pipeline.

This fixes issue #198.
See this Chromium [discussion](https://bugs.chromium.org/p/chromium/issues/detail?id=1098514) for background.

I first tried setting a completely artificial timestamp that would would tick once per coreVideoRefresh call.
That helped for cases where the encoding pipeline drops frames as it cannot keep up with the emulator.
However, it did not help when the emulator itself changes the frame rate because of lack of resources or simply because there is nothing to render.
Therefore I went for a real time based time stamp with a seed (I assume the original code had this to avoid easy to decrypt patterns in an encrypted stream) and with a 120fps resolution.

I tried a bit higher and lower resolutions and settled on this...
I am not sure how much of a difference small variations make so we can keep experimenting on them.
Also we can put this use of a timestamp behind a config flag if we find out it makes things worse for certain games...

### DIFF
--- a/pkg/emulator/libretro/nanoarch/naemulator.go
+++ b/pkg/emulator/libretro/nanoarch/naemulator.go
@@ -55,7 +55,7 @@ type constrollerState struct {
 
 // naEmulator implements CloudEmulator
 type naEmulator struct {
-	imageChannel chan<- *image.RGBA
+	imageChannel chan<- GameFrame
 	audioChannel chan<- []int16
 	inputChannel <-chan InputEvent
 
@@ -78,15 +78,20 @@ type InputEvent struct {
 	ConnID    string
 }
 
+type GameFrame struct {
+	Image     *image.RGBA
+	Timestamp uint32
+}
+
 var NAEmulator *naEmulator
 var outputImg *image.RGBA
 
 const maxPort = 8
 
 // NAEmulator implements CloudEmulator interface based on NanoArch(golang RetroArch)
-func NewNAEmulator(etype string, roomID string, inputChannel <-chan InputEvent) (*naEmulator, chan *image.RGBA, chan []int16) {
+func NewNAEmulator(etype string, roomID string, inputChannel <-chan InputEvent) (*naEmulator, chan GameFrame, chan []int16) {
 	meta := config.EmulatorConfig[etype]
-	imageChannel := make(chan *image.RGBA, 30)
+	imageChannel := make(chan GameFrame, 30)
 	audioChannel := make(chan []int16, 30)
 
 	return &naEmulator{
@@ -102,7 +107,7 @@ func NewNAEmulator(etype string, roomID string, inputChannel <-chan InputEvent) 
 }
 
 // Init initialize new RetroArch cloud emulator
-func Init(etype string, roomID string, inputChannel <-chan InputEvent) (*naEmulator, chan *image.RGBA, chan []int16) {
+func Init(etype string, roomID string, inputChannel <-chan InputEvent) (*naEmulator, chan GameFrame, chan []int16) {
 	emulator, imageChannel, audioChannel := NewNAEmulator(etype, roomID, inputChannel)
 	// Set to global NAEmulator
 	NAEmulator = emulator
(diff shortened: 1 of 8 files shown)


sergystepanov:
Looks cool. And what about Opus RTP packets? Shouldn't they be timestamped in the same way?
https://github.com/giongto35/cloud-game/blob/2703692d1a15d4e14ee2491f00b4ad65c5606384/pkg/webrtc/webrtc.go#L327-L348

88hcsif:
I have not been focusing much on the audio yet.
Probably the audio track would also benefit from similar changes.
But even if using the same timestamps I think the two tracks would still be independent.
I will give it a try and then learn a bit more how the two tracks can be synced.

sergystepanov:
Of course they are independent until mixed together on the server.
What I think (if I [understood it](https://www.cs.columbia.edu/~hgs/rtp/faq.html) right), it will be enough to share same reference clock of the system in each RTP-stream instead of `Timestamp:   r.Uint32()` + 1 and RTCP SR will (should) contain propper global-to-our timestamp mappings for both.

giongto35:
The code is LGTM but let keep it for a while if anyone wants to continue the discussion. And I think Audio is OK with the same Video approach.

I'm thinking of how to test this easily. I tried to write a packet x 100 times per an image frame with and without timestamp. 
According to chromium forum, they said there is a RTPTimestamp graph that I can track. However, my chrome chrome://webrtc-internals doesn't show that. I don't know if there is an equivalent one. @88hcsif Do you just check with the TimeFrameDelay from stat?

Btw, I took a look at https://github.com/pion/rtp/blob/master/packetizer.go#L16, there is EnableAbsSendTime, Don't know if it is also an attempt to attach timestamp. https://webrtc.googlesource.com/src/+/refs/heads/master/docs/native-code/rtp-hdrext/abs-send-time. NTP time is the proper way to sync.

sergystepanov:
It's `lastPacketReceivedTimestamp`.
image

sergystepanov:
It seems that Wireshark's RTP analyzer is [a killer](https://wiki.wireshark.org/RTP_statistics), don't know how to make it work, though. (:

88hcsif:
My understanding is far from complete but here is what I gathered so far.
Note: I think we are mostly on the same page so some or all of these remarks might be redundant.

I think it will be beneficial to add the same logic to the audio stream for the same reasons as the video stream:
- to cope with audio samples dropped in the audio encoding pipeline (I have not checked if this can happen for audio, but probably)
- to cope with reduced sample rate when the system does not call retro_run fast enough.

Even if the two streams use the same time base for the timestamp that is not enough to sync audio and video.

Instead, WebRTC uses other another layer of the protocol (RTCP) to translate video timestamps and audio timestamps into real NTP timestamps.
I will look into EnableAbsSendTime, it might be exactly what enables this extra sync messages at the RTCP level.

I mostly just check the difference between receivedFrames and decodedFrames.
I will check if there is something similar for audio.
I too struggled to find the exact stat in webrtc-internals or in https://fippo.github.io/webrtc-dump-importer/ that produces the screenshot in that Chromium discussion named "RTP timestamp".
It might be lastPacketReceivedTimestamp but it also might be something else that you can only see if you process the webrtc-internals dump with some other tool different from https://fippo.github.io/webrtc-dump-importer/...

sergystepanov:
Then it is probably [this](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpContributingSource).

giongto35:
This is a very nice discussion. I'm happy to learn a lot from you guys.

88hcsif:
Just a quick note here that:
Adding the same logic to the audio stream is not straight forward, I tried and the audio came out pretty terrible.
There is some extra complexity due to the fact that audio samples are aggregated and resampled before being fed to the Opus encoder.
And the fact that 44.1kHz or 48KHz, the timestamp probably needs to have a pretty accurate microsecond resolution.
I think we can leave the logic as it is for now.
I will still look into whether everything is set up correctly to sync audio and video at the RTCP level.

My main issue with audio so far comes from PAL games and the Mupen64plus core.
But I think it that has more to do with the core playing a PAL game as if it were a US/Japan game rather than WebRTC...

jtestard:
> There is some extra complexity due to the fact that audio samples are aggregated and resampled before being fed to the Opus encoder.

@88hcsif how about taking the timestamp once all audio frames are collected for a sample, but before encoding starts?

PR #212, opened 2020-06-28
Add tests for emulator state corruption

sergystepanov:
### Goals

- [x] Add simple tests for save states.
- [x] Prevent save states from abuse with client-side save/load calls throttling.
- [x] Investigate interesting behavior of some cores*.
- [x] [Bonus] Add simple bench metrics for the emulator and its cores to track performance between code revisions.

### Notes

I wasn't able to reproduce save state corruption with the mGBA Libretro port. The main emulator has multiple reports about corrupted states, so it's possible that's the emulator issue.
Interestingly enough, some cores like mGBA, it seems, can change their state independently from retro_run calls. For example, Nestopia never shows such behavior.
I added a number of tests for several parts of the app's emulation with some mocking examples included. The problem with tests is that the app uses a bunch of hard to isolate global vars and it's hard to track all that async close / deinit calls.
There was a bug with cloud save store with missing credentials handling (look at the code).
And the bench part and tests will be included into CI some time later.


sergystepanov:
It's better not to add useless code that don't add anything useful like this PR. So I discarding it.

giongto35:
@sergystepanov , I'm looking at it. Most of this is test right? I think it's never useless to have a test. It also helps testing not only state corruption but other things as well.

PR #173, opened 2020-04-26
Add settings module

sergystepanov:

PR #213, opened 2020-06-28
Add multitap support to SNES

88hcsif:
Change SNES emulator to Snes9X.
mednafen_snes_libretro does not seem to support multitap.
mednafen_snes_libretro was also crashing on my Pi despite coming from the Retroarch nightly builds for ARMv7.
Hopefully we are all happy to change.

Add client control to toggle multitap.
Unfortunately we cannot have multitap always enabled.
There are SNES games that only support 2 players such that if multitap is enabled player 2 stops working (for example Super Mario Kart).
We could compile a list of games that work with multitap but for now it is simpler to give control to the user.
And for now control is just a hidden keyboard shortcut, we can decide to add something to the UI in the future.

I have not tried playing any PS1 game but it would be interesting to see how multitap works for PS1 in the future.

Sneak a minor commit to fix PS3 controller mapping.


giongto35:
Hi @88hcsif,

Can I know what is multitap? Is it specific for some games or some cores only? My search result is like it is port for multiplayer?

`There are SNES games that only support 2 players such that if multitap is enabled player 2 stops working (for example Super Mario Kart).`
Does it mean 2 players will stop working if we enable multitap. So Multitap cannot always be enable?

I don't want to pollute user interaction, so I don't feel like introducing multitap as toggle. I'm Ok with some simple hacky code which enables that for specific games. We can keep the code tidier as well.

88hcsif:
It is a device peripheral available for some old consoles including [SNES](https://nintendo.fandom.com/wiki/Super_Multitap) and [PS1](https://en.wikipedia.org/wiki/PlayStation_Multitap) that allowed to expand the number of peripherals and usually players.
Some cores that emulates some of these old consoles support this type of emulated device.
The libretro API does not seem to be very specific on how these special device work, it is a bit left to the core to decide how to implement them.
Hopefully multitap on other cores work somewhat similarly.

>>There are SNES games that only support 2 players such that if multitap is enabled player 2 stops working (for example Super Mario Kart).

>Does it mean 2 players will stop working if we enable multitap. So Multitap cannot always be enable?

Yes, correct.
My guess (but I might be completely wrong) is that the core is emulating very closely what the console did in reality.
Probably multitap was something released later in the life of the SNES (I am not even sure it was released officially by Nintendo) and early games do not know how to cope with this device.
So in a game like Super Mario Kart, player 2 stops working completely if the multitap is plugged in.
But in a game like Teenage Mutant Ninja Turtle, player 2 keeps working even if the multitap is plugged in and even if the game only supports 2 players (and therefore it does not make real use of the multitap).

I agree that we should not pollute user controls.
Unfortunately it is really hard to compile a list of games that support multitap.
But we could stick to just the games we are interested on (for example bomberman is really fun with multitap support).
So if I am a developer that wants to deploy this project for my own use, it is my responsibility to keep the list up to date for the games I like.
However, if we are ever going to allow "end users" to load their own ROM, things are a bit more complicated because it becomes harder to maintain a list... name matching would probably be unreliable... but I believe N64 ROMs have some kind of unique code and maybe SNES ROMs do too.
Just for comparison, in Retroarch the user has to change the settings manually to enable multitap.
By the way, is allowing users to upload ROMs something you ever considered?

giongto35:
`By the way, is allowing users to upload ROMs something you ever considered?`
I think it's fine, personalization is a good thing to have. User can have control on ROMs list, core list, setting and game history ... Bring this to user will reduce the work on platform a lot. I think sergy is having an attempt to make a proper setting page https://github.com/giongto35/cloud-game/pull/173 and we can introduce the multi-tap setting on it.  We can have some hacky way first, I prefer we stick with the list of game we are interesting. 

Indeed, making this project universally works for all retro games and has the same functionality as RetroArch is not my goal. I would love to see its performance improve like what you help with the WebRTC lag. Besides, I want to part away with retro ultimately as well, so we no longer face issues related to LibRetro. Regarding to it, one of the ways I'm trying to do is spawning window virtual machine, running any game, and capture screen. However, I have a headless server issue you mentioned last time when graphic game is unable to run on a headless server.

88hcsif:
Sounds good I will either look into adding some sort of list or just wait for the settings menu to be ready.

sergystepanov:
@giongto35, maybe merge these changes? It's just a multitap on/off key passed to the backend.
Merged version of conflict files in the archive if you need them.
 
[deconflicted.zip](https://github.com/giongto35/cloud-game/files/5188437/deconflicted.zip)

giongto35:
Thank you @sergystepanov very much for helping with the merge conflict. : D Let me go through everything again.

giongto35 (approved):

Issue #180, opened 2020-05-03
Worker crashed/SIGSEGV while closing Sushi the cat


trichimtrich:
### Env
Dev
MacOS

### Step to produce
1. Run coordinator + connect worker
2. Run `Sushi the cat` game
3. Wait for 5 secs
4. Close the browser tab

### Expected
No crash

### Output
<details>

```
I0503 08:05:32.739968   19154 main.go:29] Initializing worker server
I0503 08:05:32.740110   19154 worker.go:47] Starting monitoring server for overwork
I0503 08:05:32.740162   19154 monitoring.go:53] Starting monitoring server at :6601
I0503 08:05:32.740169   19154 monitoring.go:67] Prometheus metric is enabled at :6601/worker/metrics
I0503 08:05:32.742265   19154 storage.go:25] Warn: Failed to create client: dialing: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
I0503 08:05:32.742323   19154 worker.go:154] Listening at port: localhost: 9000
I0503 08:05:32.742361   19154 handlers.go:95] Worker connecting to coordinator: ws://localhost:8000/wso?zone=
I0503 08:05:32.744018   19154 handlers.go:72] Connected to coordinator successfully. &{0xc00007aa40} <nil>
I0503 08:05:32.744422   19154 coordinator.go:45] Received serverID  f1b830c3-b63b-4cc5-bf87-ced43eb4d5ee
I0503 08:05:34.028183   19154 worker.go:75] &{0xc00011d900 0xc000152300 {} 0x4093440 false false false false 0xc00038c1c0 {0xc0003ba0e0 map[] false false} map[Access-Control-Allow-Origin:[*]] true 0 -1 0 false false [] 0 [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] [0 0 0 0 0 0 0 0 0 0] [0 0 0] 0xc000394070 0} echo
I0503 08:05:34.066044   19154 coordinator.go:57] Received a request to createOffer from browser via coordinator
I0503 08:05:34.066145   19154 webrtc.go:111] === StartClient ===
I0503 08:05:34.091253   19154 webrtc.go:131] Add video track
I0503 08:05:34.091313   19154 webrtc.go:142] Add audio track
I0503 08:05:34.091607   19154 webrtc.go:200] Created Offer
[...]

PR #215, opened 2020-07-01
Fix race condition when closing encoders

88hcsif:
I am applying the principle found [here](https://go101.org/article/channel-closing.html): **don't close a channel from the receiver side and don't close a channel if the channel has multiple concurrent senders**.

Media will now wait for the emulator to gracefully close, and then close the encoder input.
The encoder will wait for media to gracefully close its input, and then close its output and release any memory that cannot be garbage collected.

There are also other channels in the project that could be refactored to follow this principle.
However, they are a bit more complex and more importantly they seem to work, so let's not change it if it ain't broken.

This should fix issue #180.

### DIFF
--- a/pkg/encoder/h264encoder/encoder.go
+++ b/pkg/encoder/h264encoder/encoder.go
@@ -4,12 +4,14 @@ import (
 	"bytes"
 	"log"
 	"runtime/debug"
+	"sync"
 
 	"github.com/gen2brain/x264-go"
 	"github.com/giongto35/cloud-game/pkg/encoder"
 )
 
 const chanSize = 2
+var mu sync.Mutex
 
 // H264Encoder yuvI420 image to vp8 video
 type H264Encoder struct {
@@ -73,6 +75,9 @@ func (v *H264Encoder) startLooping() {
 		}
 	}()
 
+	mu.Lock()
+	defer mu.Unlock()
+
 	for img := range v.Input {
 		err := v.enc.Encode(img.Image)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/encoder/h264encoder/encoder.go
```diff
@@ -4,12 +4,14 @@ import (
...
 	"log"
 	"runtime/debug"
+	"sync"
 
 	"github.com/gen2brain/x264-go"
 	"github.com/giongto35/cloud-game/pkg/encoder"
 )
 
 const chanSize = 2
+var mu sync.Mutex
```
giongto35, line 14: we might have multiple encoder, so better put lock in H264Encoder

pkg/encoder/vpx-encoder/encoder.go
```diff
@@ -42,6 +43,7 @@ GoBytesType get_frame_buffer(vpx_codec_ctx_t *codec, vpx_codec_iter_t *iter) {
 import "C"
 
 const chanSize = 2
+var mu sync.Mutex
```
giongto35, line 46: similarly, put lock in VpxEncoder struct as well

giongto35:
Why you need to have mutex for controlling close order?

giongto35:
The way I check is after a session is closed, all goroutine is cleared. If there are multiple sessions join and leave, the number of goroutine is still a constant.
I also tried to follow the principle. I think there is someplace I feel difficulty to strictly follow it because the system is bidirectional. Emulator is the writer for image stream, but is reader for input. WebRTC is the reader for image stream but writer for game input. Room is the middleware.
Therefore, it's a bit messy for closing order right now, so It's great if you want to refactor it : D.

88hcsif:
Yeah it is definitely a complex architecture which needs some extra thought and that actually might be already correct as it is.

I moved the mutexes (now a channels) in the struct so that they are not global as you suggested.
I also moved the call to close the encoder input inside the stop/release function so that it should not be possible to misuse the encoder and cause a deadlock.

The reason why I was using mutexes and now channels is to coordinate the looping go routine and the stop/release function.
We cannot release memory until we know for sure that the loop is stopped.
I changed to channels because I guess they are more common and more appropriate in golang.

giongto35:
Yes, the code looks good to me now :P. Thank you for the fix

Issue #225, opened 2020-09-14
having trouble running N64 on ubuntu, No available video device


ravelab:
Am I missing a video driver? Other emulators work. Thanks

-----------------------------------
--- System audio and video info ---
-----------------------------------
  Aspect ratio:  1.3333333730697632
  Base width:  320
  Base height:  240
  Max width:  320
  Max height:  240
  Sample rate:  44100
  FPS:  60
-----------------------------------
panic: 

goroutine 17 [running, locked to thread]:
github.com/giongto35/cloud-game/pkg/emulator/libretro/nanoarch.initVideo()
        /root/cloud-game/pkg/emulator/libretro/nanoarch/nanoarch.go:376 +0x993
github.com/giongto35/cloud-game/pkg/emulator/libretro/nanoarch._cgoexpwrap_58c817539c33_initVideo()
        _cgo_gotypes.go:752 +0x25
make: *** [Makefile:77: dev.run] Error 2

giongto35:
Hi @ravelab ,
Thank you for reporting it : D. I will try checking it later.

sergystepanov:
I think this is related solely to SDL2 render route of the app if I understand the error from the message correctly: `No available video device` -- that's the SDL2 error.

It's similar to what I have inside WSL2:
![image](https://user-images.githubusercontent.com/846874/93072762-870e4300-f68a-11ea-98fd-36c4f3f7a976.png)

Guess, it needs X11 server (local: `DISPLAY=:0`, or some remote), and since n64 emulator is hardware-accelerated -- some GPU drivers (Mesa or proprietary).
@giongto35, will it work with xvfb (in mem "goodbye gpu-acceleration" mode)? Just curious, never worked with it.

(also want to mention that it didn't work for me when I tried to run initial PR, because my CPU has only ancient GL version support)

giongto35:
@sergystepanov you made some good observations here. Let me try with XVFB, I think It can solve this problem on Linux Container.

regarding to 
```
It'll be easier to debug by running glxinfo with glxgears from mesa-utils, if 'gears' are OK then n64 should work as well.

Oh, wow! WSL2 can work in direct mode:
```
I'm not totally get this point. Are you talking about running N64 in window environment with Ubuntu?

sergystepanov:
Yes, X Window, I mean one possible scenario that will help with OpenGL debugging for him if you are going to do so.
As always, there is no mention of his run environment, maybe he's willing to run it on a desktop.

[...]

PR #226, opened 2020-09-15
Add initial xvfb system dependency for OpenGL cores support in the cl…

sergystepanov:
…oud environment.

Related to #225.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,12 +35,15 @@ RUN apt-get update && apt-get install -y \
     libopusfile0 \
     libsdl2-2.0-0 \
     libgl1-mesa-glx \
+    xvfb \
   && rm -rf /var/lib/apt/lists/*
 
 COPY --from=build ${BUILD_PATH}/bin/ ./
 # it assumed there are only binary files yet
 RUN cp -s $(pwd)/* /usr/local/bin
 COPY web ./web
-COPY assets/emulator/libretro/cores/*.so ./assets/emulator/libretro/cores/
+COPY assets/emulator/libretro/cores/*.so \
+     assets/emulator/libretro/cores/*.cfg \
+     ./assets/emulator/libretro/cores/
 
 EXPOSE 8000 9000 3478/tcp 3478/udp
--- a/README.md
+++ b/README.md
@@ -3,9 +3,9 @@
 [![Build](https://github.com/giongto35/cloud-game/workflows/build/badge.svg)](https://github.com/giongto35/cloud-game/actions?query=workflow:build)
 [![Latest release](https://img.shields.io/github/v/release/giongto35/cloud-game.svg)](https://github.com/giongto35/cloud-game/releases/latest)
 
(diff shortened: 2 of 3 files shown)


sergystepanov:
While testing n64 stuff, I noticed constant 2-3 sec audio hiccups, video stream seemed OK.

giongto35:
Thank you very much. It's ok to have N64 work on WIndow unless you want to do it. Maybe Native Window has equivalent virtual monitor.
Theoretically, video should not output to monitor, I'm not sure about N64 implementation in Libretro. It makes the whole N64 implementation a bit hacky. We have a good workaround now btw.
Let me try testing the Audio, maybe the CPU is too high when running N64 that makes the Audio processing delay.

sergystepanov:
@giongto35, as far as I know, it's not possible to render anything with the OpenGL backend without some window associated with its context. That's just how it was made. 
Both n64 Libretro emulators use OpenGL. It's possible that Vulkan API or DirectX backends don't require windows, idk, need to read.
Personally, I'm not a fan of xvfb since it doesn't support hw 3D rendering and primarily used for 2D rendering in clouds (browser page render and such). Not sure whether it would be useful to have 3D acceleration for clouds (do they have GPUs at all for a sane price?), but for own streamer I'm gonna try to go in that way anyway (:. Maybe, something like [this](https://virtualgl.org/About/Introduction).

About native Windows app n64 support: it's just a problem with my laptop GPU and drivers (Intel HD2000, HD3000). With any slightly modern GPU it will run without problems, I think. Native build for sure doesn't require X Windows. Windows has windows since version 1.0 (:
And btw, why you call Windows -- Window?

giongto35:
Hi @sergystepanov ,
Interesting to know that. I wasn't aware of the 3d acceleration. You also provide information about virtual GL that will help me a lot later if I'm looking for optimization.
`And btw, why do you call Windows -- Window?`
: D And TIL Microsoft Window**s** with s.

Thank you very much. I learn many new things talking with you.

giongto35:
@sergystepanov  the coordinator issue you mention is also valid I think but seems like we can spawn multiple Xvfb and forward to different port in case. :/

sergystepanov:
@giongto35, about:
> @sergystepanov the coordinator issue you mention is also valid I think but seems like we can spawn multiple Xvfb and forward to a different port in case. :/

I thought about it a bit more and here is my theory (:
if we wanted to run multiple workers on the same machine, we could use just one xvfb-provided virtual display, because each worker instance creates its own OpenGL context including a separate rendering framebuffer for the emulator instance output.
And one fake display (in my PR it's just 1 px sized) is enough to allow of creating multiple fake windows during SDL's OpenGL context initialization since windows can overlap, stack, and whatever, also we aren't drawing anything on window surfaces.
So, one problem that I can think of: we need to init xvfb explicitly with the separate command to allow sharing its fake display among the workers.  The current `xvfb-run` method makes xvfb process child to the main command call (`worker`), therefore it'll exit when the worker exits and that shared display will be destroyed.

PR #232, opened 2020-10-01
Add game library module

sergystepanov:
This PR adds a new game library module for better app mods, coordinator/workers separation.
Now all workers need just one packet with an explicit file path in order to run a ROM, so it's the coordinator job to manage ROMs.
An experimental game library directory watcher was implemented also, it's disabled by default (coordinator app param: `-libMonitor`), which allows updating the app game list without restarting.
```
I1001 12:09:46.126017    8016 game_library.go:140] [lib] scan... started
I1001 12:09:46.127026    8016 game_library.go:261]
--------------------------------------------
--- The Library of ROMs                  ---
--------------------------------------------
    anguna (anguna.gba)
    Super Mario Bro2s (nes\Super Mario Bro2s.nes)
    psxtest (psxtest.cue)
    Super Mario Bros — копия (Super Mario Bros — копия.nes)
    Super Mario Bros (Super Mario Bros.nes)
    Sushi The Cat (Sushi The Cat.gba)
--------------------------------------------
--- ROMs: 006                   1.0095ms ---
--------------------------------------------
I1001 12:09:46.128023    8016 game_library.go:183] [lib] scan... completed
```

The game list is stored as a map instead of a slice restricting duplicate game names. That made no sense with the original slice structure since any number of duplicate ROMs (in nested folders) was shown in the list but only the first one in the slice would run for any user choice among them.
To implement it properly, it would require to break room-game name mappings and use something like hashes instead.

Waiting for v2 merge.


pkg/games/game_library.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+
+	log.Printf("[lib] scan... completed\n")
+}
+
+// watch adds the ability to rescan the entire library
+// during filesystem changes in a watched directory.
+// !to add incremental library change
+func (lib *library) watch() {
+	watcher, err := fsnotify.NewWatcher()
```
giongto35, line 191: this fsnotify is interesting library. Thanks for teaching me it.

pkg/games/game_library.go
```diff
@@ -0,0 +1,289 @@
...
+					return
+				}
+				if event.Op == fsnotify.Create || event.Op == fsnotify.Remove {
+					// !to try to add the proper file/dir add/remove scan logic
+					// which is tricky
+					repo.Scan()
+				}
+			case _, ok := <-watcher.Errors:
+				if !ok {
+					return
```
giongto35, line 212: should just continue I think

sergystepanov: I think you are right, thanks.

PR #233, opened 2020-10-01
Add anti-abuse debounce and throttle functions

sergystepanov:
Some small useful functions. Would be great to implement them on the server-side too.


giongto35:
oh, this is nice function. I never work with javascript in my profession before, so i'm really impressed with this code : v

PR #234, opened 2020-10-01
Deduplicate latency check server list

sergystepanov:
Deduplicate multi-instance worker addresses on the client and all the rejects are gonna be over 9000 ms.
Needs to be done on the server as well.


sergystepanov:
Rebased with the latest master changes.

pkg/coordinator/handlers.go
```diff
@@ -342,14 +342,18 @@ func (o *Server) findBestServerFromBrowser(workerClients map[string]*WorkerClien
...
-	workersList := []*WorkerClient{}
-	addressList := []string{}
+	var workersList []*WorkerClient
+	var addressList []string
+	uniqueAddresses := map[string]bool{}
 	latencyMap := map[*WorkerClient]int64{}
 
 	// addressList is the list of worker addresses
 	for _, workerClient := range workerClients {
+		if _, ok := uniqueAddresses[workerClient.PingServer]; !ok {
```
giongto35, line 352: I think we can use uniqueAddresses directly as set. I will fix it later : D

giongto35 (approved):

PR #237, opened 2020-10-06
Fix errors/misuse with OpenGL-based core API

sergystepanov:
This PR is dedicated to fixing some errors and misuses in the OpenGL HW-accelerated Libretro cores renderer.

How Libretro cores utilize OpenGL.

If a Libretro core needs to use OpenGL, for example, to emulate 3D-accelerated games, in that case, your Libretro frontend has to own and share the OpenGL context with the core. The only job for the core is to render graphics for the next frame and put the final result into a shared FrameBufferObject (`FBO`) [[3]](http://www.songho.ca/opengl/gl_fbo.html#overview), all that done with a special render-to-texture OpenGL extension that should be supported by the video driver (OpenGL 3+). After that, your frontend can read the image from the `FBO` and show it in any convenient way.

A simplified list of steps to deal with OpenGL Libretro cores:
-  Initialize the OpenGL context.
   - Make a window with the help of SDL. And we will need a window because OpenGL context initialization needs to gather data from the windowing system even though the rendering is going to be off-screen. (_perhaps it is possible to reuse an existing window in the OS or even destroy new window right after context creation but this idea requires verification_) `¯\_(ツ)_/¯`
   - Using some OpenGL Loading Library (e.g. go-gl [Glow]), load pointers to OpenGL functions at runtime, core, and extensions.
   - Call the OpenGL init function and that will be enough.
   - Don't forget that before calling any GL functions in a thread, context should be made current (thread-local variable). That way each thread gonna have its own copy of the OpenGL context state to change and should flush these changes during your next context-make-current calls or will keep them inside the thread otherwise [[4](https://docs.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-wglmakecurrent)]. 
- Initialize FrameBufferObject and a texture to draw on.
   - By default windowing system provide you with its own framebuffer but we need to create one explicitly in order to minimize data copy and increase the performance for core off-screen rendering.
   - To let OpenGL perform rendering into a texture, an image of a texture object must be attached to a framebuffer (`GL_COLOR_ATTACHMENT`).
   - Additionally, you can specify the depth and stencil components of the framebuffer (all three also are called attachments) if the core returns info about the usage of those.
   - Bind FBO's name to the current context for reading/writing operation allowance with `GL_FRAMEBUFFER` param (_I'm not sure about the need for this step since it makes no sense here without any data to read, I yet to know how to properly bind/unbind that buffer between frontend/core, especially with an amazing ability of go-routines to jump between threads_).
- Notify the core that you are done (call `context_reset` callback).
- Grab the image from the FBO in the `video_refresh` callback when it's ready.
   - Check the `RETRO_HW_FRAMEBUFFER_VALID` flag as an indicator for HW rendering.
   - Bind the framebuffer for reading op.
   - Read pixels from it (`glReadPixels` or `glTexSubImage2D`).

Sources:
1. Hans-Kristian Arntzen Implementing a Hardware Accelerated Libretro Core ([pdf](https://github.com/libretro/libretro.github.com/blob/master/documents/libretro-gl.pdf))
2. Interactive Computer Graphics. 6.1 OpenGL Version, Context, and Profile ([link](https://sites.google.com/site/gsucomputergraphics/educational/initialization-tasks-in-an-opengl-program/opengl-context-and-profile))
3. songho.ca OpenGL ([link](http://www.songho.ca/opengl/gl_fbo.html#overview))
4. WinAPI wglMakeCurrent ([link](https://docs.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-wglmakecurrent))
5. The Red Book (:<

Issues:
- SDL2 is a pretty bloated framework just for window creation, GLFW is much lighter
- [x] to check OpenGL Cores Libretro [doc](https://docs.libretro.com/development/cores/opengl-cores/)
```
Video driver should support render-to-texture extension (OpenGL 3.0+ Core).
```
- [x] unnecessary OpenGL 4.1 Core profile use, it's better to follow [2.1 profile](https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/)
```
Switched to OpenGL 2.1 bindings.
```
- [x] OpenGL coordinate system is flipped, use of slow and unnecessary third-party dependency just for the image vertical flip was unjustified

https://github.com/giongto35/cloud-game/blob/a42aebec886f479be43ad33661c43077085b1c6b/pkg/emulator/libretro/nanoarch/nanoarch.go#L153-L166
> 12. OpenGL's Lower Left Origin
> 
> Given a sheet of paper, people write from the top of the page to the bottom. The origin for writing text is at the upper left-hand margin of the page (at least in European languages). However, if you were to ask any decent math student to plot a few points on an X-Y graph, the origin would certainly be at the lower left-hand corner of the graph. Most 2D rendering APIs mimic writers and use a 2D coordinate system where the origin is in the upper left-hand corner of the screen or window (at least by default). On the other hand, 3D rendering APIs adopt the mathematically minded convention and assume a lower left-hand origin for their 3D coordinate systems.
> 
> If you are used to 2D graphics APIs, this difference of origin location can trip you up. When you specify 2D coordinates in OpenGL, they are generally based on a lower left-hand coordinate system. Keep this in mind when using glViewport, glScissor, glRasterPos2i, glBitmap, glTexCoord2f, glReadPixels, glCopyPixels, glCopyTexImage2D, glCopyTexSubImage2D, gluOrtho2D, and related routines.
> 
> Another common pitfall related to 2D rendering APIs having an upper left-hand coordinate system is that 2D image file formats start the image at the top scan line, not the bottom scan line. OpenGL assumes images start at the bottom scan line by default. If you do need to flip an image when rendering, you can use glPixelZoom(1,-1) to flip the image in the Y direction. Note that you can also flip the image in the X direction. Figure 4 demonstrates using glPixelZoom to flip an image.
> 
> Note that glPixelZoom only works when rasterizing image rectangles with glDrawPixels or glCopyPixels. It does not work with glBitmap or glReadPixels. Unfortunately, OpenGL does not provide an efficient way to read an image from the frame buffer into memory starting with the top scan line.

Rewrite it either similarly to `rotationFn` aka 0-cost frame slice reader (+ combine with rotations somehow) or use GL matrix transform (which is slower):

> Libretro treats top-left as origin. OpenGL treats bottom-left as origin. To be compatible with the libretro model, top-left semantics are preserved. Rendering normally will cause the image to be flipped vertically. To avoid this, simply scale the final projection matrix by [1,− 1 , 1 ,1].

```
Added a function that flips Y coordinate for each pixel after rotation and before RGBA image draw.
```

- [x] Pixel formats can change from core to core, misused for textures and the frame buffer (BRG/RGB)
```
Used corresponding OpenGL pixel formats and types based on core callback values.
```

- FBO read speed can be increased with [PBO](http://www.songho.ca/opengl/gl_pbo.html).

- [x] macOS "main thread [problem](https://discourse.glfw.org/t/multithreading-glfw/573/5)" solution causes a crash with the native Windows build of the app (https://github.com/golang/go/wiki/LockOSThread)

From go-sdl2 FAQ:
> Why does my program crash randomly or hang? Putting runtime.LockOSThread() at the start of your main() usually solves the problem (see SDL2 FAQ about multi-threading).
> 
> UPDATE: Recent update added a call queue system where you can put thread-sensitive code and have it called synchronously on the same OS thread. See the render_queue or render_goroutines examples from https://github.com/veandco/go-sdl2-examples to see how it works.

```
Added SDL/GL functions into main.thread wrapper. (rewrite later: remove mainthread lib 
in favor of go-sdl2's included render-queue)
```
- [x] huge memory leak on Windows, Linux with [mupen64plus-libretro-nx](https://github.com/libretro/mupen64plus-libretro-nx) core

image
```
Seems that some problem with retro_load_game / retro_unload_game function in the worker.
SDL/OpenGL deinitialization broken at least in Windows, a bunch of zombie? threads remain.
Postponed until backporting normal test runners for Libretro cores.

[won't fix] RetroArch with the same core has this issue. 
```
- [x] to try to check strangely deterministic audio stutter with the native Windows build :( WebRTC audio jitter all over the place
image
```
A possible cause is in the Mesa drivers (slow software llvmpipe renderer).
Disabling OpenGL drawing in the core removes audio stuttering.
```


giongto35:
:O WOW
Looks really interesting, I haven't got time to understand more about it. Thank you very much for your initiative.

sergystepanov:
@giongto35, I opened process explorer just for a quick check on leaks for n64 games, and look at the mem graph:

image

+500 megs unfreed memory for every game run, on the first run it allocates 800 Mb (: That's ridiculous. I'm checking on Windows, later will recheck for Linux, maybe that's specific for Win, but I think that it's not.
Guess, we need to check for leaks from time to time also.

sergystepanov:
The leak definitely either in C or OpenGL part. ):

sergystepanov:
Really like to debug segfaults (: I thought that the addition of OpenGL stuff will rise the complexity with the arithmetic progression, but it's more like geometric, especially for multi os compatibility.

image

sergystepanov:
To go further I need to backport my Libretro core run tests, that better to do with new PR.

giongto35:
Sorry, I haven't got time to look at it now. it looks like a simple PR but you give very detailed follow-up that I also want to dive in. :D

sergystepanov:
@giongto35, we're not in the rush. Actually, I'll port and add old [PR](https://github.com/giongto35/cloud-game/pull/212/files) core tester into this PR then, that's gonna be just ~300 lines, so not that much.
One thing -- that macOS main thread stuff testing: theoretically I know roughly how it should work but I'm not entirely sure until verify it with my own eyes. Geez, I have to read about and install a macOS Virtualbox image because of that. :(

sergystepanov:
With a fresh install of RetroArch it's leaking too now.

image

giongto35:
I just went through it. So the "simplified list of steps for making it work" you mentioned is the way Libretro does behind the scene right? This PR is trying to Fix some misuses with openGL in our code.
Is it good to be merged now? I see it pretty big

sergystepanov:
I meant: "A simplified list of steps to deal with OpenGL Libretro cores".
Initially, I was planning to fix some minor issues but turns out there were a little bit more problems with the code.
I'm not done yet.

sergystepanov:
Tried to run OpenGL core (n64) with Github CI but it's failing on Ubuntu during SDL window creation:
```
Could not create GL context: BadValue
```

sergystepanov:
Added CI artifact with core-rendered images, like:

image

sergystepanov:
Setting the core profile for GPU-less machines won't allow creating an SDL window context.

https://github.com/giongto35/cloud-game/blob/ea0a433ee982d9e687154c39ce7a8ea857403d69/pkg/emulator/libretro/nanoarch/nanoarch.go#L381-L400

Gotta rewrite it. Github CI result without it:

linux-n64-0x903d7ba8

- [x] Ubuntu CI
- [ ] Windows CI -- to add MESA 20 manually, + annoying manual config
- [ ] macOS CI -- probably needs Quartz, some drivers

sergystepanov:
go-gl substantially increases compilation speed.

sergystepanov:
@giongto35, I think, it's ready to be merged now.

Summary:
- Fixed OpenGL cores so they run on all 3 OSes.
- Added special tests for Github CI runners to check emulators with eyes (:.
Every PR or push now generates an [artifact](https://github.com/giongto35/cloud-game/actions/runs/335607007) with test frames from the 3 most used emulators including bloody n64 (software OpenGL since Github runners don't have GPU). Example frames in the artifact or up above. The main idea was to check if someone broke the emulation and when it happened. I made this frame rendering step optional, so it won't fail the whole CI run.
- Refactored some code and started to separate graphics stuff.
- About that memory leak. It needs a lot of time for debugging the mupen core itself. I wasn't able to check it properly with Valgrind, because the Linux library is compiled without debugging info. I don't have time to mess with recompilation. RetroArch has the same memory leak (1.9.0 + the latest Mupen core).

giongto35 (approved):

giongto35:
Hi, Sorry for the delay. I'm having some deadlines at work.
It's a really good piece of Code. Thanks for setting up all the tests and CI. I will try to relook later when I have time. The OpenGL stuff is complicated though :-<

PR #245, opened 2020-11-03
Add initial new audio

sergystepanov:


giongto35:
It looks interesting? what is it going to do?

sergystepanov:
The main reason is to remove my old messy code (to add a new messier one (:), maybe make it compatible with various backends such as GStreamer, to try to implement a canonical resampler (yours works almost all the time, actually) or use SoX, and to try one experimental thing with varying audio frame duration of WebRTC Opus. Not to wait for 20ms of audio from the emulator, just send ASAP all the audio stretched to the nearest allowed by Opus audio frame duration. That's also could be related to the old Firefox bug with audio distortion of not 20ms audio frames.

sergystepanov:
The original resampler is broken for 2.5,5,40...ms frame durations.

PR #249, opened 2020-11-29
Add new Libretro core manager module

sergystepanov:
**The first step** is to make workers and coordinators configurable externally in order to allow changing app params at the runtime without recompilation.
So the default `configs/config.yaml` file was added. It's shared between workers and coordinator. The apps will be searching for it in:
-  The current directory
-  The `configs` dir from where the app was started
-  The `{home}/{user}/.cr` directory

[`kkyr/fig`](https://github.com/kkyr/fig) is used as a backend lib for configuration handling.
Config read chain: YAML file, then some params can be overridden with app runtime flags, overriding with env vars is not implemented yet but can be supported.

**The second step** is to add core downloader functionality. Package manager.

Repository.
Supported: raw zip archive, Buildbot, GitHub (raw-mode). The current Libretro nightly repository (Buildbot) contains zipped libs only.

Downloader.
There are two ways of dealing with the non-monolithic nature of the app (or replicated service):
let the coordinator handle downloads for all the workers (1), 
let them handle everything by themselves as a closed box in isolation (2).

(1) There are some problems associated with the app in terms of a distributed system:
- It needs a single source of truth.
- It needs a mutual agreement between all its nodes (consensus)

Coordinator `(C)` to Worker `(W)` relation is 1:N, `(C)` to Storage `(S)` and `(W)` to `(S)` is 1:1.
```txt
                      (Storage)
         |--- (W) --------|
(C) ---- |--- (W) --------|
 |       |--- (W) --------|
 |________________________|
```

(2) Or just try a decentralized way of handling core downloading.
```txt
               __magic__
               |       |
          |------ (W) ------|
(Repo) ---|------ (W) ------|--- (Storage)
          |------ (W) ------|
```

Current flow:
- when a worker starts it blocks after cws connection until it verifies that cores are in place (downloaded, it can be disabled in the config)
- get current architecture (core lib extension)
  - if the predefined mapping between system architectures and dynalib extensions doesn't contain such arch then it will fallback to greedy loop (try to load whatever loads)
- for each core defined in the config, check
  - if a core file exists -- then skip
  - else build download URL for nonexistent core-libs on the disk
- push the list of URLs into the core-downloader baked by [`cavaliercoder/grab`](https://github.com/cavaliercoder/grab)
- if a repository contains archived cores -- unpack and delete all archives
- Profit

About concurrent workers on the same machine: it's handled by an IPC file locking mechanics [`gofrs/flock`](https://github.com/gofrs/flock) lib, same as mutex the shared lock file is stored in the system temp directory visible to all workers. Should work on mac, nix. win.

I decided left some cores (mac n64) in this source because they were manually compiled, buildbot repo doesn't contain them and I'm using them on mac N64 OpenGL CI tests (: Probably, will add some fallback mirror repo when downloading (new GitHub project), to search there not found in buildbot cores.


giongto35:
So between 2 ways, what do you pick?

sergystepanov:
2nd, since it's easier to keep that abstraction in your brain (:
It's already usable, just needs multiprocess locking (like yarn.lock metafile) that will protect from redownloads inside one dir and readiness check + let coordinator know when workers are ready to play the game.

giongto35:
I think you mentioned 1. because I proposed something like keeping the coordinator as a "source of truth"?
Indeed, I'm toward the 2nd as well, let the workers sync with the config, and just download what cores do not exist. I think it's good and simple solution. Just the "list of cores"/ config is fetched from coordinator instead of repo. But either way is fine.

And when you mention "repo" here, do you mean it's Github repo?

sergystepanov:
> I think you mentioned 1. because I proposed something like keeping the coordinator as a "source of truth"?

I can vaguely remember what we talked about but if you mean that terminology then yes -- after your talk, and no, I mentioned it at all because I considered coordinator as the only super-node that can order an action to a worker and expect some reaction. Workers' only job is to download ordered files. All those multiprocess problems are handled by the coordinator through his inner network connection to the workers. It's more fun to implement but harder and time-consuming. So I choose the 2nd option.

> Indeed, I'm toward the 2nd as well, let the workers sync with the config, and just download what cores do not exist. I think it's good and simple solution. Just the "list of cores"/ config is fetched from coordinator instead of repo. But either way is fine.

If you so want it (:, to push worker config from coordinator what essentially an on-the-fly config change, I can try to add that too.

> And when you mention "repo" here, do you mean it's Github repo?

Where? Repository? Here repository is a place where cores are stored and will be downloaded by workers. Either buildbot Libretro, or, actually, a partial mirror on GitHub in some different repo (take a look, it can be accessed without auth with HTTP like that)
https://github.com/giongto35/cloud-game/blob/723a3273740055963ca0fc2b3a8f37af8b477778/pkg/downloader/downloader_test.go#L5-L11

sergystepanov:
Now I'm thinking cores and games should have some manager interface (CURD) with a separate metadata file outside the config file like go.mod controlled only with CLI interface, and some backend page on the site to control them by admin. That will be cool. 
hashtag #ihavenothingbettertodowithmytime2021 (:

giongto35:
> Now I'm thinking cores and games should have some manager interface (CURD) with a separate metadata file outside the config file like go.mod controlled only with CLI interface (or finally use etcd), and some backend page on the site to control them by admin. That will be cool.
> hashtag #ihavenothingbettertodowithmytime2021 (:

oh didn't think of that but it's really nice idea! Thinking about that, manager interface can be the interface of the repo we mentioned above. Repo can be backed by Git. It will be a seamless experience now.

giongto35:
> Where? Repository? Here repository is a place where cores are stored and will be downloaded by workers. Either buildbot Libretro, or, actually, a partial mirror on GitHub in some different repo (take a look, it can be accessed without auth with HTTP like that)
Your current proposal is good enough if you do it in this PR.
The interface is nice to have because it's totally abstract away the repo (HTTP to Github looks hacky ). In term of software design, with the interface, we can change the underlying repo (like storing in DBMS) and decouple from Github.

sergystepanov:
Sure.

Test video.

https://user-images.githubusercontent.com/846874/102724362-19745180-4320-11eb-9f76-64fd0f4c24b3.mp4

giongto35:
> Sure.
> 
> Test video.
> 
>  annoy02.mp4

Cool :+1: , why name it annoy02.mp4 :p

sergystepanov:
It should be fast since it's a multithreaded download, 100Mbit, and now they pack cores into zip -- only 15 MB are those cores on the video (75MB unpacked).
(annoy is the last prefix of videos in the recording program, it just increments a counter on your next recordings, too lazy to change)

sergystepanov:
@giongto35, I'm almost finished with this PR, so you could make a quick code review if you want (: I need to update the first post with general code edits overview and to test a bit more.

giongto35:
Hi Sergy, Is it good to be reviewed and pushed now : D

configs/config.yaml
```diff
@@ -0,0 +1,161 @@
+#
```
giongto35, line 1: Awesome : D. you can not go wrong with yaml

configs/config.yaml
```diff
@@ -0,0 +1,161 @@
...
+    httpsPort: 443
+    httpsKey:
+    httpsChain:
+
+worker:
+  network:
+    # a coordinator address to connect to
+    coordinatorAddress: localhost:8000
+    # ISO Alpha-2 country code to group workers by zones
+    zone:
```
giongto35, line 50: this `zone` and `httpsKey`, `httpsChain` is only for production right? I can try new deployment with these flags?

pkg/emulator/libretro/repo/github/repository.go
```diff
@@ -0,0 +1,20 @@
...
+	"github.com/giongto35/cloud-game/v2/pkg/emulator/libretro/repo"
+	"github.com/giongto35/cloud-game/v2/pkg/emulator/libretro/repo/buildbot"
+)
+
+type Repo struct {
+	buildbot.Repo
+}
+
+func NewGithubRepo(address string, compression string) Repo {
+	return Repo{Repo: buildbot.NewBuildbotRepo(address, compression)}
```
giongto35, line 14: the github repo is now mocked by buildbot ?

pkg/emulator/libretro/manager/remotehttp/manager.go
```diff
@@ -0,0 +1,96 @@
...
+		fmu:    flock.New(fileLock),
+	}
+}
+
+func (m Manager) Sync() error {
+	declared := m.Conf.GetCores()
+	dir := m.Conf.GetCoresStorePath()
+
+	// IPC lock if multiple worker processes on the same machine
+	m.fmu.Lock()
```
giongto35, line 60: Cool!

pkg/webrtc/ice.go
```diff
@@ -0,0 +1,62 @@
...
+		Username:   user,
+		Credential: credential,
+	}
+}
+
+func ToJson(iceServers []webrtc.IceServer, replacements ...Replacement) string {
+	var sb strings.Builder
+
+	n := len(replacements)
+	serversN := len(iceServers)
```
giongto35, line 32: Look like this can be rewritten better :/ 
We can use strings.Join 2 times, one to build a formatted version of each iceServer, and then join all iceServers together

pkg/emulator/libretro/repo/github/repository.go
sergystepanov: Yes. By that, I meant exact same file structure as in buildbot. It's a mirror with free GitHub hosting (:

sergystepanov: I'll create one later.

pkg/emulator/libretro/manager/remotehttp/manager.go
sergystepanov: More cooler that it works at least in win and nix (: (didn't know that Windows has filelocks)

pkg/webrtc/ice.go
sergystepanov: This' a superfast version (: We can just json.Marshal(iceServers) + loop replacements (:

pkg/emulator/libretro/repo/github/repository.go
sergystepanov: If you think about that, it's free, versioned, tagged (with branches), replicated (with forks) (:

configs/config.yaml
sergystepanov: I didn't change any related logic, so yes, it's for your DO.
It will be great if you try it but I forgot to add environment vars into that for convenient use with Docker.
The chain of config params should be: file -> (almost done, next PR -- config requested from coordinator if no file) -> environment vars -> comandline arguments.

sergystepanov:
> Hi Sergy, Really nice work!! Is it good to be reviewed and pushed now : D
> Really nice re-written of config management. I like this yaml config style.

Thanks (: I left some replies to your review comments. I forgot to add config param override from the system environment (for Docker). I'll merge after adding that and it'll be ready to try.

pkg/webrtc/ice.go
sergystepanov: Need to test what's faster. Essentially, strings.join on iceServers will do exactly the same logic. I should rewrite strings.Join on iceServer params (: and bench.

pkg/emulator/libretro/repo/github/repository.go
giongto35: it's ok for now, I think let's keep buildbot. I was wondering if repo can be some storage like s3, or the google cloud storage like currently.

sergystepanov: You mean with a client (login:pass). Haven't worked yet with cloud storage providers but it's easy to add now. In any case, it needs some background management (copy buildbot).
I assume it's related more to games repo which is yet to be implemented. For that, yes, we will need private storage.

pkg/webrtc/ice.go
sergystepanov:
```
goos: windows
goarch: amd64
pkg: github.com/giongto35/cloud-game/v2/pkg/webrtc
BenchmarkIces
BenchmarkIces/toJsonOriginal
BenchmarkIces/toJsonOriginal-4         	  989529	      1171 ns/op	     520 B/op	       9 allocs/op
BenchmarkIces/toJsonJoins
BenchmarkIces/toJsonJoins-4            	  480134	      2430 ns/op	    1120 B/op	      27 allocs/op
BenchmarkIces/toJsonMarshal
BenchmarkIces/toJsonMarshal-4          	  599916	      1933 ns/op	     864 B/op	       5 allocs/op
PASS
```
As expected, my implementation without joins is the fastest (original is ~1800 ns/op), joins is the slowest, and simple marshaling 2nd (:

PR #254, opened 2021-01-03
Migrate to pion/webrtc v3

sergystepanov:
Since pion/webrtc v3 was [released](https://github.com/pion/webrtc/wiki/Release-WebRTC@v3.0.0).

TBD.


sergystepanov:
@giongto35, it seems that I was able to port it almost painless and it's working.
As you may remember, for our video with variable framerate we decided to generate and inject our own timestamps into outgoing packets. Unfortunately, pion/webrtc/v3 doesn't export the RTP packetizer anymore, so we can't inject timestamps directly into the local track structure that we are using. It's possible that new [Interceptors](https://github.com/pion/interceptor/blob/master/README.md) functionality could be helpful there but I didn't try it, instead, I copied the track struct from the v3 and replaced writeSample method with the code similar to what we have before.
And Trickle ICE seems to be working just fine.

giongto35:
Thanks Sergy very much, Another great work!
Don't know the reason why Pion no longer supports it, it makes me doubt if the packetizer with timestamp helpful? Last time, when we introduce the timestamp, I still feel it didn't help much. Did you test and see it work.

giongto35:
btw, the PR is good to go. Pion V3 signature looks nicer!

sergystepanov:
> Don't know the reason why Pion no longer supports it, it makes me doubt if the packetizer with timestamp helpful? Last time, when we introduce the timestamp, I still feel it didn't help much. Did you test and see it work.

I didn't test it, since I thought you have handled it, and that problem initially was reproduced on N64 Mupen core, ARMv7 (32bit) Raspberry Pi, which I used to have but lost somewhere (now I have ARMv8 RPi only). In theory, it should help when a core drops fps below the provided constant (60). There is another option based on the duration of the rendered video chunk for which we don't have all needed Libretro callbacks implemented (:
Anyway, for cloud-morph you can use v3 standard localTrack with a constant duration ([1/your target fps] ms) for the sample, or if you can measure and pass the true value of variable video sample duration somehow that will be cool.

giongto35:
Yes, I definitely port cloud-morph to pion v3. I'm copying a lot of your code to cloud-morph. it always work very nice. : D
btw, there is a trending pion webrtc project today, you may find it interesting. https://news.ycombinator.com/item?id=25627673 .

sergystepanov:
@giongto35, thanks, checked it. Looks cool for a toy project. Wouldn't recommend adopting such coding technics, though (:

PR #259, opened 2021-01-07
Fix Docker run for bridged Docker network in Windows, macOS

sergystepanov:
During the ICE gathering step of WebRTC, it searches for available network interfaces with some random ports.
Therefore we couldn't use the default port export config with the bridge network in Docker.
If we explicitly define a port range for ICE candidates and swap internal Docker IP onto the host IP, it should work.
There are new ice-related params in the config that undefined by default so the app runs as usual. These params are set with env in the new docker-compose script.


pkg/webrtc/connection.go
```diff
@@ -19,6 +20,25 @@ func NewInterceptedPeerConnection(conf Configuration, interceptors []interceptor
 		i.Add(itc)
 	}
 
-	api := NewAPI(WithMediaEngine(m), WithInterceptorRegistry(i))
-	return api.NewPeerConnection(conf)
+	settingEngine := SettingEngine{}
+	if conf.IcePorts.Min > 0 && conf.IcePorts.Max > 0 {
+		if err := settingEngine.SetEphemeralUDPPortRange(conf.IcePorts.Min, conf.IcePorts.Max); err != nil {
```
giongto35, line 25: @_@ wow really cool

docker-compose.yml
```diff
@@ -9,13 +9,20 @@ services:
     environment:
       - DISPLAY=:99
       - MESA_GL_VERSION_OVERRIDE=3.3
+      - CLOUD_GAME_WEBRTC_ICEPORTS_MIN=50000
```
giongto35, line 12: Is there any risk if the range is this limited? Don't know if there is any recommended range

giongto35 (commented):
It looks good. It's still weird to me that host network worked before and stop working now. But this change is good (y).

docker-compose.yml
sergystepanov: No idea. It can be easily reconfigured changed.

sergystepanov:
> It looks good. It's still weird to me that host network worked before and stop working now. But this change is good (y).

It's possible that you're was using the old Docker Toolbox based on VirtualBox which should support bridge on all OSes.

I tried Windows WSL2 Linux and it's not working with 127.0.0.1 because it's not it's address (you should use what in eth0).

Could you check this Docker on native Linux and Mac? And then I'll be merging into master.

sergystepanov:
Actually, we need as much as possible ports open, since any spectator watching a game will need a port from that range.

sergystepanov:
Tested max number of opened (forwarded) Docker ports on Windows and it seems it works up until 999 opened ports, after that the app just won't respond. (:

giongto35:
Oh I pulled master, and run. And it's totally fine. : D
image
Maybe it's the issue on some machines. We can remove "The app runs as a Docker container only in Linux. "

sergystepanov:
@giongto35, interesting. So the "host" mode works then. Can you confirm that you are running Docker Desktop?
Ok, I'll revert config changes leaving config options.

PR #282, opened 2021-02-21
Fix x264 build trash

sergystepanov:
Added frame color encoding switch specifically for libx264 library versions < v159. On these versions, it will try to use the old value if it can determine the installed x264 build version.
This hack is meant for modern Debian/Ubuntu distros since they still use library versions less than v160. Otherwise, just install library version 160 or newer.


sergystepanov:
@giongto35, should work on this branch with versions 150+. Need some testing on 160+ but still, it's a hack.

sergystepanov:
Tested on Ubuntu 20.04 (v155) and 20.10 (v160) and it seems to be working.

PR #285, opened 2021-02-23
Refactor and update VPX codec

sergystepanov:
### Done
- Removed a lot of unused C code.
- Only VP8 related code is preserved.
- Added VP8 codec configuration to the config file.
- Added a generic video encoder pipe for all used video codecs.

### Issues

- VP9 doesn't work in the current master and after the refactoring.


sergystepanov:
@giongto35, I just looked at Mario with VP8 encoding and it seems the same h264 encoding artifacts are present. They are literary the same. wtf. Even worse, maybe. If both codecs do so, then I think these artifacts could be related to the YUV color space, chroma luma stuff. But I can't recall that I had these artifacts before. Maybe I just didn't pay attention to it.
In any case, if they share the same artifacts, it's probably better to use h264 as the default codec then.
I'll try to tweak it after the rewrite...

feature007

sergystepanov:
Yep. It's YUV artifacts. That's YUV 4:2:0 converter either trash by itself or 4:2:0 is not meant for small images.
Here RGB, YUV raw export.

image


RGB
download (7)
YUV
download (5)

pkg/encoder/pipe.go
```diff
@@ -0,0 +1,69 @@
...
+
+	size := int(float32(vp.w*vp.h) * 1.5)
+	yuv := make([]byte, size, size)
+
+	for img := range vp.Input {
+		// if buffer is full skip frame
+		// why the hell we need that?
+		//if len(vp.Output) >= cap(vp.Output) {
+		//	continue
+		//}
```
sergystepanov, line 51: @giongto35, need help with this part of the code. Could you please explain its purpose if you remember? You had it in one of the encoders.

sergystepanov: Right now we have 1000/60fps * 20chanLen = 333ms window until it can be overrun. Even in the "worst" case if the framerate is constant, normally it's less than 60fps, that's a lot of time.

giongto35: you mean why I convert from RGBA to YUV? Because the encoder accept YUV, so I did that

sergystepanov: @giongto35, I mean:
```
if len(vp.Output) >= cap(vp.Output) {
  continue
}
```
Makes no sense.

sergystepanov: I'm pretty sure it never worked. I assume you wanted to skip encoding when the channel overflows. As is it makes no sense as I said on the output channel, should that be on the input channel instead?

PR #287, opened 2021-03-03
Add new YUV converter with avg color calc

sergystepanov:


pkg/encoder/yuv/yuv.go
```diff
@@ -0,0 +1,137 @@
...
+	for i := 0; i < yuv.threads; i++ {
+		pos, hh := C.int(yuv.w*i*yuv.chunk), C.int(yuv.chunk)
+		// we need to know how many pixels left
+		// if the image can't be divided evenly
+		// between all the threads
+		if i == yuv.threads-1 {
+			hh = C.int(yuv.h - i*yuv.chunk)
+		}
+		go func() { defer wg.Done(); C.luma(yuv.dst, src, pos, yuv.ww, hh) }()
+		go func() {
```
giongto35, line 130: It also costs some goroutine creation per frame here same like thread.
What I recommend is to do something like threadpool. You spawn goroutine up most before process any frames, each goroutine will be in charge of a predefined area.
Then for each frame, we just notify each goroutine to process their corresponding area.

After that, we can go one step further by omitting wg.Wait. We make each area is processed separately in async manner. Once a frame finishes processing, then we push out the frame

For example:
We have a frame 4x4, we spawn 4 goroutines, each one is in charge of 2x2
Game controller get to frame 4th
go routine 1: 2x2 finished 3 frames
go routine 2: 2x2 finished 2 frames
go routine 3: 2x2 finished 1 frames
go routine 4: 2x2 finished 3 frames

Once goroutine 3 finished, => it sends a completed first frame. and so one

PR #289, opened 2021-03-07
Fix Pion WebRTC codecs

sergystepanov:
Since Pion's webrtc v3.0.13 and this [fix](https://github.com/pion/webrtc/commit/e5c8c659caa915041dd6dbcaacbb85fe28733820) if both peers have some fmtp set then codecs should match exactly. By default Pion defines a list of supported codecs with fmtp set but different browsers set their own fmtp values by default as well therefore they may not match.
To fix this we can define our own list of codecs on the server with empty fmtp so it will match any client codecs.

See: https://github.com/pion/webrtc/blob/e5c8c659caa915041dd6dbcaacbb85fe28733820/rtpcodec.go#L99


sergystepanov:
For some reason, if you define multiple RTPCodecParameters for one MimeType (i.e. video/H264) with different PayloadTypes and empty SDPFmtpLines, it won't add anything from this list into the codecs struct for the search (see [the first message](https://github.com/giongto35/cloud-game/pull/289#issue-586310145)).
Right now, for h264 some payloads have fmtp specified and some that used in Chrome, Firefox doesn't have.
Fix later.
Tested in Chrome/Firefox on vpx, h264, opus.

PR #299, opened 2021-03-19
Remove jQuery

sergystepanov:


sergystepanov:
@giongto35, just a quick overview of the changes:
- Successfully removed jQuery, tested both desktop and mobile Cr/FF.
  > I think, now the code is basically identical to the old one. Found some strange logic in the touch input controller regarding button handlers. It needs to be revamped but in another PR, maybe.
- Removed some legacy (needed for IE) CSS properties.
- Removed fullscreen scale of the interface on Android because it's ugly without some padding, also that Github ribbon at the top right corner goes right on top of the load/save buttons.
- Can't help myself but I tweaked D-pad dashes on the PNG image since they are horribly misaligned, buttons are next. Do you have by any chance Photoshop or Gimp source files for the interface elements? That would be helpful.

giongto35:
Hi Sergey, Really thanks for this because I keep hearing that we can move away with jquery but the source code has some jquery and I don't know what is the equivalent vanilla js.
Regarding to the misalignment, how does it look like now. Can you capture the screen or I can check out this branch and see?

sergystepanov:
Looks like this:

image

I made them the same size and centered. (I see my mistake on a shadow with the bottom dash because the light source is at the top left corner, so it should be as on the original, will fix it. Also, horizontal dashes should have shadow too, at the top (: )
Most of the buttons are misaligned too (: That's not a problem but it's annoying for a person with OCD (:.

giongto35:
lol, it takes me few minutes to see the difference. Im not OCD person
Yes, actually we should make it aligned too.
The psd is here: https://drive.google.com/file/d/1R_XOmfNlYuenwRJO6SNnf4DaY-BhSrB5/view?usp=sharing . I outsource the UI : D but I can help fix the UI when I have time.

I can checkout this branch and see the current UI right?

sergystepanov:
Cool! With the source .psd I might come back to my old attempt of redoing everything in raw CSS (:
Yeah, you can check this branch with my alignments (or open these images on Github).

sergystepanov:
It seems that .psd is the "dark" (green) version, do you have the current light version? Or is it the source that you recolored?

sergystepanov:
Interesting. I think I found the root cause of the old issue with the high-frequency tap (touch interface).
Take a look at the video where the circle pad with my new click handler and buttons have their old handler.

https://user-images.githubusercontent.com/846874/112490535-802a6080-8d90-11eb-983a-d50d63e19338.mov

It seems that when we handle touch events (start, stop), they fire both very quickly therefore input polling function (every 10ms) can't register a tap. As an experimental workaround, I added a small delay before touchstop event on the circle pad and it seems it's working much better even though it's not a really good implementation.
+1 bug to fix later (:

sergystepanov:
Anyway, I think I'm done with testing this PR, so you can check it out if you want, will be merging it soon.

PR #312, opened 2021-04-08
Use GLFW instead of SDL2

sergystepanov:
GLFW is much lighter and simpler just for our purposes of GL window context creation.

- [x] Linux -- good
- [x] Windows -- good
- [ ] macOS -- ugly

As of version 3.3, GLFW has some problems with macOS, the documentation says:
```
How do I create an OpenGL 3.0+ context?
The only OpenGL 3.x and 4.x contexts currently supported by macOS are forward-compatible, core profile contexts.
The supported versions are 3.2 on 10.7 Lion and 3.3 and 4.1 on 10.9 Mavericks.
In all cases, your GPU needs to support the specified OpenGL version for context creation to succeed.

To create either an OpenGL 3.2 or later context, you should set the following hints:
...
This will give you a context for the highest version of OpenGL greater than or equal to 3.2 
that is supported by your OS and GPU. This works because no features were deprecated between 
OpenGL 3.2 and 4.1, making forward-compatible OpenGL 4.1 backards-compatible with 3.2. 
If the requested version is not supported, window creation will fail.

If your code requires OpenGL 3.3 or 4.1, you can specify those versions instead, 
as well as any valid OpenGL version in between.
```
This means that CI builds (headless server) won't work without a compat profile or some X11 hacking:
```
Unable to create basic Accelerated OpenGL renderer.
Core Image is now using the software OpenGL renderer. This will be slow.
panic: VersionUnavailable: NSGL: The targeted version of macOS only supports 
forward-compatible core profile contexts for OpenGL 3.2 and above
```

This PR will be postponed until better times and new GLFW versions.


sergystepanov:
Archive.

PR #320, opened 2021-05-27
CD Flow to redeploy after push

giongto35:


.github/workflows/redeploy/config.yaml
```diff
@@ -0,0 +1,192 @@
+# Application configuration file
```
sergystepanov, line 1: Don't forget to remove this file if you decide to use environment variables instead.

.github/workflows/redeploy/redeploy_specific.sh
```diff
@@ -0,0 +1,34 @@
...
+do
+    echo $ip_address
+    ssh-keyscan -H $ip_address >> ~/.ssh/known_hosts
+    sleep 2
+
+    if [ "$ip_address" == "167.172.70.98" ] || [ "$ip_address" == "cloudretro.io" ] 
+    then
+        launchcommand="coordinator > /tmp/startup.log"
+    else
+        launchcommand="Xvfb :99 & worker --coordinatorhost cloudretro.io --zone \$zone > /tmp/startup.log"
```
sergystepanov, line 15: xvfb is used only for n64 OpenGL, so you might remove it in case if there are no n64 games on your hosts.

.github/workflows/redeploy/redeploy_specific.sh
```diff
@@ -0,0 +1,34 @@
...
+    ssh root@$ip_address "mkdir -p /cloud-game/configs"
+    rsync ./.github/workflows/redeploy/config.yaml root@$ip_address:/cloud-game/configs/config.yaml
+    run_content="'#! /bin/bash
+    echo $PASSWORD | docker login https://docker.pkg.github.com --username $USERNAME --password-stdin; 
+    ufw disable;
+    docker system prune -f;
+    source /etc/profile; 
+    docker pull docker.pkg.github.com/giongto35/cloud-game/cloud-game:latest;
+    docker rm cloud-game -f;
+    docker run --privileged -d --network=host --env DISPLAY=:99 --env MESA_GL_VERSION_OVERRIDE=3.3 -v cores:/usr/local/share/cloud-game/assets/cores -v /cloud-game/configs:/usr/local/share/cloud-game/configs -v /cloud-game/games:/usr/local/share/cloud-game/assets/games -v /cloud-game/cache:/usr/local/share/cloud-game/assets/cache --name cloud-game docker.pkg.github.com/giongto35/cloud-game/cloud-game bash -c \"$launchcommand\"'"
```
sergystepanov, line 27: Might split long lines this: https://github.com/giongto35/cloud-game/blob/master/scripts/install.sh#L11-L16

sergystepanov (approved):
Something like this. LGTM.

.github/workflows/redeploy/config.yaml
giongto35: oh, I think I will use this approach. Updating env for each machine is more tedious to me and put everything in config.yml makes it more intuitive to other users.

.github/workflows/redeploy/redeploy_specific.sh
giongto35: ah that's right, I can update like that

.github/workflows/redeploy/config.yaml
sergystepanov: @giongto35, I meant that it's better to override a couple of config params by defining and passing environment vars into the Bash script and then inside Docker. The reason behind that: we're removing config file copying or modifying work which we have to do now every time we change default config.yaml in the app, that file plays the role of the app config structure initializatior. Internal config structure is initialized with empty values (0, '', nil) without it, which will cause a crash. So, we better leave the default config.yaml inside the Docker image and just override needed values externally with env vars. We can do that in your Bash script in the repo here.
I'll try to fix this later.

.github/workflows/redeploy/redeploy_specific.sh
sergystepanov: Ok, will change if you won't.

.github/workflows/redeploy/config.yaml
sergystepanov: Yes, I expect that the default config.yaml will be changed quite often.
You use Docker image or build with the default config inside and just override needed params with env variables, flags, whatever. That's how I see it. In your example just set system env CLOUD_GAME_ENVIROMENT=prod on your server.
And yes, it is better to keep all the config params together not scattered all around the whole codebase.
How do you see it?

PR #326, opened 2021-07-03
Remove port binding for docker config

tianrenz2:
Docker-compose does not support host mode along with the port binding anymore. Additionally there is no necessity for using port binding under current configuration since it takes no effect.


sergystepanov:
Thanks. This will make the code a little bit clear.

PR #327, opened 2021-07-04
Track worker's user count.

sergystepanov:
If a worker is shared, then without tracking players it's impossible to tell when it's free, so we add a simple (not thread safe, meh) user counter.


giongto35:
oh don't know why I did like that : D This would fix the issue.
Don't know if IsAvailability is necessary because I have this cleanWorker based on room count : https://github.com/giongto35/cloud-game/blob/747d0d702ca19e9e82c9a8e25d5cd841f45feca3/pkg/coordinator/handlers.go#L376

PR #328, opened 2021-07-05
Track connected user count of workers

sergystepanov:
If a worker (game) is shared, then without tracking players it's impossible to tell when it becomes available for a new game which leads to emulator share and crashes.

### DIFF
--- a/pkg/coordinator/handlers.go
+++ b/pkg/coordinator/handlers.go
@@ -217,7 +217,9 @@ func (o *Server) WS(w http.ResponseWriter, r *http.Request) {
 
 	// Assign available worker to browserClient
 	bc.WorkerID = wc.WorkerID
-	wc.IsAvailable = false
+
+	wc.ChangeUserQuantityBy(1)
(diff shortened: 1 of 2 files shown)


pkg/coordinator/handlers.go
```diff
@@ -217,7 +217,9 @@ func (o *Server) WS(w http.ResponseWriter, r *http.Request) {
 
 	// Assign available worker to browserClient
 	bc.WorkerID = wc.WorkerID
-	wc.IsAvailable = false
+
+	wc.ChangeUserQuantityBy(1)
```
giongto35, line 221: btw, you can try this from golang : D. It supports atomic computation. https://golang.org/pkg/sync/atomic/

sergystepanov: I don't like its syntax: readandgetputInt345. Also, CAS instructions can be quite tricky. Already used it before in the cloud-game code, doesn't matter performance-wise for this chunk of code.

PR #337, opened 2021-07-30
Update server HTTPS configuration

sergystepanov:
The HTTPS server configuration has been fixed and reworked.
 -   New configuration params:
     ```yaml
     server:
       address: :8000
       https: false
       # Letsencrypt or self cert config
       tls:
         address: :443
         # the public domain of the coordinator
         domain: localhost:8000 
         letsencryptUrl:
         # if both are set then will use certs
         # and Letsencryt instead
         httpsKey:
         httpsCert:
     ```
  - The `server.[tls.]address` params now accept a string in the following format `host:port`, instead of a single port value.
     ```yaml
     // before
     server.address: 8080 -> :8080
     // after
     server.address: :8080 -> :8080
     server.address: local:8080 -> local:8080
     ```
  - A secure HTTPS/WSS connection can be enabled now with the `server.https` config param and there are two possible ways of using it. The first one, when you leave both `server.tls.httpsKey` and `server.tls.httpsCert` params empty, in that case, you will be using Letsencrypt auto certification config, its default URL can be changed with the `server.tls.letsencryptUrl`. Alternatively, you can specify HTTPS cert and key files, if you want, as an example, test HTTPS with a local self-signed certificate.

- Worker pushes its zone and public ping server address (builds automatically) params during internal connection to the coordinator.

An example of SSL/TLS config:
```yaml
# ...
coordinator:
  # ...
  # HTTP(S) server config
  server:
    # ...
    https: true
    # Letsencrypt or self cert config
    tls:
      address: :443
      # the public domain of the coordinator
      domain: localhost:8000
      letsencryptUrl:
      # if both are set then will use certs
      # and Letsencryt instead
      httpsCert: localhost.crt
      httpsKey: localhost.key
# ...
worker:
  network:
    # ...
    # make coordinator connection secure (wss)
    secure: true
  # ...  
  server:
    address: :9000
    https: true
    tls:
      address: :444
      # LetsEncrypt config
      # if empty will use URL from Go
      letsencryptUrl:
      # Own certs config
      httpsCert: localhost.crt
      httpsKey: localhost.key
# ...
```
- Auto host+port generation code was reworked. The coordinator won't roll any ports. Workers will roll http/s ports, including the monitoring server.


cmd/coordinator/main.go
```diff
@@ -4,53 +4,37 @@ import (
...
 	"github.com/golang/glog"
 	flag "github.com/spf13/pflag"
 )
 
 var Version = ""
 
-func main() {
+func init() {
 	rand.Seed(time.Now().UTC().UnixNano())
+}
```
sergystepanov, line 21: Do we really need a random seed?

pkg/config/worker/config.go
```diff
@@ -19,34 +19,31 @@ type Config struct {
...
 }
 
 // allows custom config path
 var configPath string
 
 func NewConfig() (conf Config) {
-	if err := config.LoadConfig(&conf, configPath); err == nil {
-		conf.Loaded = true
-	}
+	_ = config.LoadConfig(&conf, configPath)
```
sergystepanov, line 42: I removed the worker's no-config option when it would pause initialization and request config from the coordinator because this stuff won't be useful.

pkg/coordinator/handlers.go
```diff
@@ -16,14 +14,15 @@ import (
...
+	"github.com/giongto35/cloud-game/v2/pkg/service"
 	"github.com/giongto35/cloud-game/v2/pkg/util"
 	"github.com/gofrs/uuid"
 	"github.com/gorilla/websocket"
 )
 
-const index = "./web/index.html"
-
 type Server struct {
+	service.Service
```
sergystepanov, line 24: Not sure if explicit interface embedding is ok.

pkg/coordinator/handlers.go
```diff
@@ -110,17 +80,19 @@ func (s *Server) WSO(w http.ResponseWriter, r *http.Request) {
...
-	wc.Println("Address:", address)
-	// Zone of the worker
-	zone := r.URL.Query().Get("zone")
-	wc.Printf("Is public: %v zone: %v", util.IsPublicIP(address), zone)
-
-	pingServer := s.getPingServer(zone)
+	public := util.IsPublicIP(address)
 
-	wc.Printf("Set ping server address: %s", pingServer)
+	connRt, err := GetConnectionRequest(r.URL.Query().Get("data"))
```
sergystepanov, line 85: All the stuff now comes directly from workers.

pkg/coordinator/internalhandlers.go
```diff
@@ -1,25 +1,32 @@
 package coordinator
 
 import (
+	"encoding/base64"
+	"encoding/json"
 	"log"
 
-	"github.com/giongto35/cloud-game/v2/pkg/config/worker"
+	api2 "github.com/giongto35/cloud-game/v2/pkg/api"
```
sergystepanov, line 8: Hate namespaces.

pkg/monitoring/monitoring.go
```diff
@@ -3,104 +3,109 @@ package monitoring
...
+func (m *Monitoring) GetProfilingAddress() string {
+	return m.server.GetProtocol() + "://" + m.server.Addr + m.conf.URLPrefix + debugEndpoint
+}
 
-	if len(conf.URLPrefix) > 0 {
-		conf.URLPrefix = strings.TrimSpace(conf.URLPrefix)
-		if !strings.HasPrefix(conf.URLPrefix, "/") {
-			conf.URLPrefix = "/" + conf.URLPrefix
-		}
+func (m *Monitoring) printInfo() {
```
sergystepanov, line 86: Don't ask why I added this function (;

pkg/monitoring/monitoring.go
```diff
@@ -3,104 +3,109 @@ package monitoring
...
+	server *httpx.Server
 }
 
-func NewServerMonitoring(cfg monitoring.ServerMonitoringConfig, tag string) *ServerMonitoring {
-	return &ServerMonitoring{cfg: validate(&cfg), tag: tag}
+// New creates new monitoring service.
+// The tag param specifies owner label for logs.
+func New(conf monitoring.Config, baseAddr string, tag string) *Monitoring {
+	serv, err := httpx.NewServer(
+		net.JoinHostPort(baseAddr, strconv.Itoa(conf.Port)),
```
sergystepanov, line 35: I want to mention that right now when the main http/s listener is created, its address (host:port, the port can be auto-generated [:0] or rolled [port+1]) is passed into monitoring or whatever needs it.
Here we merge the actual host value from the main listener with a monitoring port. Port can be rolled in worker/coordinator. The resulting address will be printed into the console.

pkg/network/httpx/address.go
```diff
@@ -0,0 +1,41 @@
+package httpx
+
+import (
+	"net"
+	"strconv"
+)
+
+// buildAddress joins network host from the first param,
+// zone from the second, and
+// the port value of a listener from the second param.
```
sergystepanov, line 10: third

pkg/network/httpx/listener.go
```diff
@@ -0,0 +1,79 @@
...
+	"syscall"
+)
+
+const listenAttempts = 42
+
+type Listener struct {
+	net.Listener
+}
+
+func NewListener(address string, withNextFreePort bool) (*Listener, error) {
```
sergystepanov, line 18: Http/s servers use listeners now just for purpose of getting an actual address (listener port).

pkg/service/service.go
```diff
@@ -0,0 +1,44 @@
+package service
```
sergystepanov, line 1: This abstraction is used for defining a container for some services (mostly HTTP servers) that automatically controls their lifecycle (start/stop).

sergystepanov (commented):
@giongto35, check this out. It's ready to be merged, I think.
Fixed auto-port roll and now it works as you intended to make it work with nested HTTP servers (redirect, monitoring).
Didn't test Letsencrypt.

giongto35:
Thanks very much. Let me check

giongto35 (approved):
Tested and worked nice, You can merge at anytime

pkg/config/shared/config.go
```diff
@@ -8,17 +8,31 @@ import (
...
-	Port       int
-	HttpsPort  int
-	HttpsKey   string
-	HttpsChain string
+	Address string
+	Https   bool
+	Tls     Tls
+}
+
+type Tls struct {
```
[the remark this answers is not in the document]
sergystepanov: It's actually not used anywhere (; Made it an anonymous struct.

pkg/coordinator/internalhandlers.go
sergystepanov: Moved into cws.api.coordinator. I ported it from "refactor netcode" PR where I removed cws package, here it's better to do as I did.

pkg/network/httpx/server.go
```diff
@@ -0,0 +1,136 @@
...
+	server := &Server{
+		Server: http.Server{
+			Addr:         address,
+			IdleTimeout:  opts.IdleTimeout,
+			ReadTimeout:  opts.ReadTimeout,
+			WriteTimeout: opts.WriteTimeout,
+		},
+		opts: *opts,
+	}
+	// (╯°□°)╯︵ ┻━┻
```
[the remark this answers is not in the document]
sergystepanov: Because of this dubious part: https://github.com/giongto35/cloud-game/blob/f28e766da7be3f4bf581b4befb5dc3e9920c1f84/pkg/network/httpx/server.go#L128-L131

I have to pass the server instance into the external HTTP handler definition callback.

PR #351, opened 2021-10-08
Create CODE_OF-CONDUCT.md

loic5:
Hello I propose to ad a code of conduct :)


sergystepanov:
Thanks for your proposal but I think there is no need for this in the project. (;

